PR #4, opened 2026-05-15
feat: Add Atkey-indexed versions of `PFunctor` as `IPFunctor`

dtumad:
# Motivation

Several places want a polynomial functor whose available shapes — and the continuation type of each shape — depend on an ambient state. Examples:

- A multi-phase adversary game where the available oracle queries change once the adversary calls the challenge oracle.
- A two-party protocol where, after each move, control passes to the other party, so the shapes available depend on whose turn it is.
- A session-typed channel where each send/receive advances the session, gating the next set of legal actions.

Plain `PFunctor` cannot express this — its `A` / `B` are flat — and its free monad is unconditionally a `Monad` only because any pure leaf can be followed by any continuation. This PR adds the indexed (Atkey-style) generalization in its natural two-index form, then specializes the free-monad layer to the endomorphic case:

```lean
structure IPFunctor (I : Type uI) (J : Type uJ) where
  A   : J → Type uA
  B   : (j : J) → A j → Type uB
  src : (j : J) → (a : A j) → B j a → I

abbrev IPFunctor.Endo (I : Type uI) : Type _ := IPFunctor I I
```

Categorically this is the polynomial diagram `I ← E → B → J`, denoting a functor `(I → Type) → (J → Type)`. `A j` is the set of shapes at output index `j`, `B j a` the response type for shape `a`, and `src j a b` the input index of the chosen response. Folding the state away (the `[Unique J]` case) recovers an ordinary `PFunctor`.

The free monad and its specializations live on the endomorphic case `Endo I`, just as ordinary monads arise from endofunctors.

# Summary

### `Basic.lean`

- `IPFunctor I J` as above, with `Obj : IPFunctor I J → (I → Type) → (J → Type)`, `CoeFun`, `Zero`, `One`, `Inhabited`.
- `abbrev Endo I := IPFunctor I I` — the endomorphic specialization that carries the free-monad layer.
- `IPFunctor.toPFunctor` — fold to a plain `PFunctor` by fixing `j := default` (requires `[Unique J]`). The earlier `[Inhabited J]` form silently discarded all non-default fibers; `[Unique J]` is the correct constraint.
- `IPFunctor.sigmaPFunctor` — alternate Σ-bundled fold (`A := Σ j, P.A j`); works for any index types. Used by `toSigmaFreeM` below.
- `IPFunctor.comp` — composition of indexed polynomials over compatible index types.
- `IPFunctor.DeterministicTransitions` — class for functors whose `src j a b` is independent of `b`. Records the unique source index `next j a` and the witness equation. Drives `Notation/Deterministic.lean`.

### `Free/Family.lean` — primitive `IPFunctor.IFreeM`

The shared indexed-family free monad over an endomorphic `IPFunctor`:

```lean
inductive IPFunctor.IFreeM (P : Endo I) (X : I → Type v) : I → Type _
  | pure {s : I} (x : X s) : IFreeM P X s
  | roll {s : I} (a : P.A s)
      (r : (b : P.B s a) → IFreeM P X (P.src s a b)) : IFreeM P X s
```

Leaves carry values from a *state-indexed* family `X : I → Type`. `IFreeM` itself is not a `Monad` — its bind is family-polymorphic in the leaf state, since different branches end at different leaf states:

```lean
protected def bind :
    {s : I} → IFreeM P X s
      → ({s' : I} → X s' → IFreeM P Y s') → IFreeM P Y s
```

Both user-facing free monads below are specializations of `IFreeM`:

- `FreeM P s α ≃ IFreeM P (fun _ => α) s` — constant return family.
- `FreeM₂ P s t α ≃ IFreeM P (fun u => PSigma (fun _ : u = t => α)) s` — terminal-state-marker family.

This shared primitive is what the second review asked for: the relationship between the two user-facing constructions is now structural rather than incidental.

### `Free/Basic.lean` — single-index `IPFunctor.FreeM`

```lean
inductive IPFunctor.FreeM (P : Endo I) : I → Type v → Type _
  | pure (s : I) (x : α) : FreeM P s α
  | roll (s : I) (x : P.A s)
      (r : (b : P.B s x) → FreeM P (P.src s x b) α) : FreeM P s α
```

Because the source index of each `roll` branch can depend on the response, `IPFunctor.FreeM P s α` is **not** a `Monad` — its bind takes a state-polymorphic continuation:

```lean
protected def bind :
    {s : I} → FreeM P s α
      → ((s' : I) → α → FreeM P s' β) → FreeM P s β
```

API parity with `PFunctor.FreeM`:

- `lift` / `liftA` lifting the base functor.
- `bind` (state-polymorphic) plus the simp family `bind_pure` / `bind_roll` / `bind_lift`.
- `bindLiftA` — specialized bind for a single `liftA` step under `DeterministicTransitions`; the continuation lands at the concrete source index `det.next s a`, no universal quantifier. Tagged `bindLiftA_eq` `@[simp]` so `simp` unfolds it without manual `unfold`.
- `Functor` / `LawfulFunctor` plus `map_pure` / `map_roll` / `map_lift`.
- `Pure (FreeM P s)` instance (the constructor wrapped at fixed `s`).
- Injectivity: `pure_inj`, `roll_inj`.
- State-indexed `inductionOn` (Prop motive) and `construct` (Type motive), plus `construct_pure` / `construct_roll` / `construct_lift`.
- `mapM` interpreting into an arbitrary monad, with the matching simp family.
- `erase : IPFunctor.FreeM P s α → PFunctor.FreeM P.toPFunctor α` under `[Unique I]`, plus `erase_punit_*` simp lemmas where transports collapse definitionally.
- `toSigmaFreeM : FreeM P s α → PFunctor.FreeM P.sigmaPFunctor α` — an *erasure* into ordinary polynomial trees that records each node's source state in the position but does not enforce the transition invariant on children. This is a forgetful map, not the main conceptual reduction; that role is now played by `IFreeM`.

### `Free/Indexed.lean` — two-index `IPFunctor.FreeM₂`

```lean
inductive IPFunctor.FreeM₂ (P : Endo I) : I → I → Type v → Type _
  | pure {s} (x : α) : FreeM₂ P s s α
  | roll {s t} (a : P.A s)
      (r : (b : P.B s a) → FreeM₂ P (P.src s a b) t α) : FreeM₂ P s t α
```

Same shape as `IPFunctor.FreeM`, but every leaf is required to sit at a single source state `t`. Strictly more restrictive than `IPFunctor.FreeM`, but in return its bind chains pre/post-indices positionally:

```lean
protected def bind : FreeM₂ P s t α → (α → FreeM₂ P t u β) → FreeM₂ P s u β
```

Carries `Pure (FreeM₂ P s s)` plus `IndexedMonad` / `LawfulIndexedMonad` (Atkey) instances, and `Functor` / `LawfulFunctor` at fixed `s t`. Mirrors the single-index API: `bind_pure` / `bind_roll`, `map_pure` / `map_roll`, `pure_inj`, `roll_inj`, `inductionOn`, `mapM`. The forgetful coercion `FreeM₂.toFreeM` embeds back into `FreeM` when the value needs to be handed to a `FreeM`-shaped API.

## `PolyFun/IPFunctor/Notation/`

Three parallel `@[doElem_elab]` overrides plugging into Lean 4.29's extensible do-elaborator. All three opt in via

```lean
set_option backward.do.legacy false
```

and gate on the expected monad type — non-`IPFunctor` monads in the same file continue to use the builtin elaborator via `throwUnsupportedSyntax`.

### `Notation.lean` — single-index `IPFunctor.FreeM`

Elaborates ordinary `do { let x ← e; … }` to `IPFunctor.FreeM.bind`-trees, threading a fresh source-index metavariable through each step. Because the bind continuation is state-polymorphic, the *tail* of the block must typecheck for every post-state — in practice this works for `pure`/`return`-terminated chains and `if`/`match` over `pure` arms.

Two custom diagnostics:

- **State mismatch** — when `let x ← e` is written at a step whose pre-state doesn't unify with the previous step's post-state. The message names both states.
- **Non-polymorphic remainder** — when the rest of the block forces the fresh post-state metavariable to a concrete value. The message asks the user to pattern-match on the previous response or write `IPFunctor.FreeM.bind` explicitly.

### `Notation/Indexed.lean` — `IPFunctor.FreeM₂`

Bind continuations statically track pre/post indices, so chains of any length compose. Uses a two-attempt elaboration strategy (state-preserving `FreeM₂ P s s` first, then a fresh post-state metavariable) so `Pure (FreeM₂ P s s)` resolves eagerly for terminal `pure x` / `return x` without postponing instance synthesis past the end of the do-block.

### `Notation/Deterministic.lean` — single-index `FreeM` + `DeterministicTransitions P`

When a step reduces to `FreeM.liftA s a` and a `DeterministicTransitions P` instance is in scope, specializes the continuation's pre-state to the concrete `det.next s a` via `FreeM.bindLiftA`. This lifts the universal-quantification constraint of the base `Notation.lean` for single-action steps. Non-`liftA` shapes (e.g. nested `do`-blocks) fall through to the generic single-index elaborator.

Imports `Notation.lean` so the deterministic override registers *after* the generic one — Lean's keyed-declaration attribute tries most-recently-registered entries first, so the specialization is checked before falling through.

### `Notation/Mixed.lean`

Cohabitation regression test exercising all three flavors in one compilation unit. Any silent drift in override priority (e.g. an import reordering that moves the `FreeM₂` override behind `FreeM`'s) surfaces here rather than at downstream call sites.

Each of the three notation files also ends with an `erase`-interop section against a `PUnit`-indexed demo `IPFunctor`, where the `erase_punit_*` / `toFreeM_*` simp lemmas collapse erased do-trees into the corresponding `PFunctor.FreeM.roll` / `pure` constructions via `rfl` or `simp`.

## `PolyFun/IPFunctor/Examples.lean`

Worked-examples module compiling the same two-phase protocol three ways:

1. `IPFunctor.FreeM₂` with pre/post-state tracked statically in the type.
2. `IPFunctor.FreeM` plus a `DeterministicTransitions` instance.
3. The result mapped down to plain `PFunctor.FreeM` via the Σ-bundled `toSigmaFreeM` — `Phase` is not `Unique`, so `erase` doesn't apply.

`Examples.Deterministic.run` and `Examples.TwoIndex.run` both carry `rfl` verification. A final example exercises `PFunctor.FreeM.equivW_of_isEmpty` to confirm `FreeM P α ≃ P.W` when `α` is empty.

### `Equiv/Basic.lean` — `IPFunctor.Equiv P Q`

Structural equivalence between `P Q : IPFunctor I J`: fiberwise `Equiv`s on `A` and `B`, plus the source-index preservation law

```lean
src_eq : ∀ j a b, P.src j a b = Q.src j (equivA j a) (equivB j a b)
```

This is equality of index *values* in `I`, not of types; in concrete cases it discharges by `rfl`. Carries `refl` / `symm` / `trans`, mirroring `PFunctor.Equiv`.

### `Lens/Basic.lean` — `IPFunctor.Lens P Q`

Cartesian morphism: forward on positions, *backward* on responses, with the contravariant `src_eq`:

```lean
src_eq : ∀ j a d, P.src j a (toFunB j a d) = Q.src j (toFunA j a) d
```

Carries `id` / `comp` plus `Lens.Equiv` with `refl` / `symm` / `trans`. Richer monoidal infrastructure from `PFunctor.Lens` is intentionally not mirrored yet — extend on demand.

### `Chart/Basic.lean` — `IPFunctor.Chart P Q`

Covariant dual of `Lens`: forward on both positions and responses, with `src_eq` in the chart direction. Same scaffolding (`id` / `comp` / `Chart.Equiv` with `refl` / `symm` / `trans`).

## Supporting edits

- `PolyFun/Control/Monad/Indexed.lean` (new) — Atkey `IndexedMonad` / `LawfulIndexedMonad` classes, with the trivial-`Unit` instance recovering every ordinary `Monad`. Consumed by `FreeM₂`.
- `PolyFun/Control/Monad/Free.lean` — tags `monadLift_eq_lift` and the `mapM_*` family with `grind =`, adds `construct_lift` and `mapM_lift`, and relocates the bundled-hom `mapM'` into its own section.
- `PolyFun/PFunctor/Free/Basic.lean` — adds `FreeM.toW` / `FreeM.ofW` between `FreeM P α` and `P.W` when `α` is empty, packaged as `equivW_of_isEmpty : FreeM P α ≃ P.W`.
- `PolyFun/Interaction/UC/OpenProcess.lean` — removes the unused `OpenNodeContext.boundaryTrace` and `OpenStep.boundaryTrace` helpers (structural trace extraction belongs downstream, not in the open-process layer).
- `AGENTS.md`, `docs/wiki/` — agent-facing tours `ipfunctor.md` and `ipfunctor-do-notation.md`; `repo-map.md` updated for the new subtree.


quangvdao:
very interesting! i'll take a look just to sanity check

quangvdao (changes requested):
Requested change: I think this PR should factor the new `IPFunctor` free constructions through a more primitive indexed-family free monad, rather than presenting `FreeM` and `FreeM₂` as parallel bespoke constructions.

The deeper object seems to be the polynomial endofunctor on indexed families `X : I → Type`:

```lean
P X s = Σ a : P.A s, (b : P.B s a) → X (P.st s a b)
```

and its free construction:

```lean
inductive IFree (P : IPFunctor I) (X : I → Type v) : I → Type _
  | pure {s} : X s → IFree P X s
  | roll {s} (a : P.A s) :
      ((b : P.B s a) → IFree P X (P.st s a b)) → IFree P X s
```

Then the current APIs should be recoverable as specializations:

```lean
IPFunctor.FreeM P s α      ≃ IFree P (fun _ => α) s
IPFunctor.FreeM₂ P s t α   ≃ IFree P (fun u => (u = t) × α) s   -- or an equivalent terminal-state encoding
PFunctor.FreeM Q α         ≃ the I = Unit case
```

This would make the relationship to ordinary `PFunctor.FreeM` much clearer: ordinary `PFunctor` is the one-state case of the indexed-family story, while `sigmaPFunctor` is only a forgetful erasure into ordinary polynomial trees. In particular, `toSigmaFreeM` records node states but does not enforce the transition invariant on children, so it should be presented as an erasure/well-formedness-forgetting map, not as the main conceptual reduction.

I think the current PR has the right ingredients, but it would be structurally stronger if `FreeM` and `FreeM₂` were derived from this shared indexed-family free construction, with the existing names kept as the user-facing special cases if desired.

Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

quangvdao (changes requested):
Additional requested changes / design direction.

The earlier request to factor through an indexed-family free construction points to an even more basic generalization: the indexed polynomial itself should probably allow distinct input and output index types. Categorically, this is the polynomial diagram

```text
I <- E -> B -> J
```

which denotes a functor between family categories

```lean
(I → Type) → (J → Type)
```

In container form, this would be something like:

```lean
structure ... (I J : Type u) where
  A   : J → Type v
  B   : (j : J) → A j → Type w
  src : (j : J) → (a : A j) → B j a → I
```

with action

```lean
P.Obj X j = Σ a : P.A j, (b : P.B j a) → X (P.src j a b)
```

The current one-index `IPFunctor I` is then the endomorphic case, where input and output indices coincide. The free constructions only make sense in that endomorphic specialization, just as ordinary monads arise from endofunctors:

```lean
P : (I → Type) → (I → Type)
```

So the requested direction is:

1. Introduce the general two-index indexed-polynomial object.
2. Define an endomorphic specialization, e.g. an abbrev like `Endo I := ... I I`.
3. Put `Free`, `FreeM`, `FreeM₂`, `IndexedMonad`, and the `do` notation only on the endomorphic specialization.
4. Do this as a full cutover of the current one-index design, rather than adding a separate generalized layer next to it.

I do not want to make the exact Lean name binding here. Reasonable candidates include `IPFunctor I J`, `IndexedPFunctor I J`, or `PFunctor.Indexed I J`; please choose whatever fits the surrounding namespace best. The important part is the two-index object plus explicit endomorphic specialization.

A rough split of what should generalize cleanly:

- General `I J`: fields `A`, `B`, `src`; `Obj` on indexed families; `CoeFun`; `Zero`; constant/unit polynomial; `sigmaPFunctor`-style erasure to ordinary `PFunctor`; deterministic transitions as `next : (j : J) → A j → I`; composition of indexed polynomials; and lenses/morphisms between indexed polynomials.
- Endomorphic only: `Free` / `FreeM` / `FreeM₂`; `IndexedMonad`; bind/do-notation; deterministic `FreeM` notation; tree `mapM`; and erasure from free trees to ordinary `PFunctor.FreeM`.

One caution: indexed lenses will need a source-index preservation law. For two indexed polynomials over the same `I J`, a lens-like morphism should roughly include

```lean
shape : ∀ j, P.A j → Q.A j
dir   : ∀ j a, Q.B j (shape j a) → P.B j a
src_eq : ∀ j a d,
  P.src j a (dir j a d) = Q.src j (shape j a) d
```

This is equality of index values in `I`, not equality of types directly, but it will induce transports in object maps because children live in fibers `X (src ...)`. In definitional cases this should often be `rfl`, but please check that the resulting lens API is still ergonomic in Lean before expanding it too far. The endomorphic case does not remove this compatibility requirement for lenses; it only makes the free-monad recursion possible.

Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

quangvdao (changes requested):
Follow-up requested changes after auditing the latest commits.

The latest push substantially addresses the earlier design requests: `IPFunctor I J`, `IPFunctor.Endo I`, the primitive `IFreeM`, and the basic indexed lens/chart/equiv structures are now in place. CI is green, and this is much closer to the intended categorical story. I still see a few changes worth making before approval:

1. Tighten or split `toPFunctor` in the general `I J` setting.

The current `toPFunctor` requires `[Unique J]`, which collapses the output index, but it does not collapse the input/source index `I`. Since `src : ... → I` can still carry real information, saying that this recovers an ordinary `PFunctor` without loss is conceptually too strong.

Please either require both indices to be unique:

```lean
def toPFunctor [Unique I] [Unique J] (P : IPFunctor I J) : PFunctor := ...
```

or split the API so the fiber-erasure and the truly collapsed case are distinct, e.g.

```lean
def fiberPFunctor (P : IPFunctor I J) (j : J) : PFunctor := ...
def toPFunctor [Unique I] [Unique J] (P : IPFunctor I J) : PFunctor := ...
```

The current `[Unique J]` operation is still useful as a selected output fiber / source-forgetting view, but it should not be documented as the ordinary-`PFunctor` recovery unless `I` is also trivial.

2. Fix the copyright header in `PolyFun/Control/Monad/Indexed.lean`.

Repo policy says Lean files should use:

```lean
Copyright (c) 2026 PolyFun Contributors. All rights reserved.
```

That file currently says `Copyright (c) 2026 Quang Dao. All rights reserved.` Please update it to the project-standard copyright line while preserving the `Authors:` line as appropriate.

3. Restore `FreeM₂.pure_inj` or remove the PR-body/API claim.

The PR body says `FreeM₂` mirrors the single-index API and includes `pure_inj`, but the current `PolyFun/IPFunctor/Free/Indexed.lean` appears to provide `roll_inj` without `pure_inj`. Please either restore a `FreeM₂.pure_inj` lemma, which seems preferable for parity, or update the PR body/docs to avoid claiming it exists.

4. Follow-up categorical surface: add identity and composition laws for indexed polynomial composition.

Now that `IPFunctor.comp` is present, it would be valuable to include the categorical identity polynomial for `IPFunctor I I` and basic left/right identity and associativity laws, probably up to `IPFunctor.Equiv` rather than definitional equality if needed. This does not need to mirror the whole `PFunctor` category stack immediately, but adding the identity and comp laws would make the new two-index interpretation much more self-contained.

Posted by Cursor assistant (model: GPT-5) on behalf of the user (Quang Dao) with approval.

quangvdao (approved):

quangvdao:
@dtumad forgot if you needed approval to merge, just did now

dtumad:
I think I just did because there was requested changes? Either way linter was also broken this whole time

PR #9, opened 2026-06-26
feat: dynamical systems as dependent lenses

dtumad:
Adds a `PFunctor/Dynamical/` layer modeling dynamical systems as dependent
lenses, following Niu–Spivak, *Polynomial Functors: A Mathematical Theory of
Interaction* (Ch. 4). A `p`-dynamical system is a lens `selfMonomial State ⟹ p`:
the interface `p` gives the positions the system can present and the directions
it can be driven by at each position, and the lens is the transition law. Because
a system is a lens, the existing `PFunctor.Lens` combinator library applies
directly instead of being re-derived for state machines. Moore machines and
deterministic automata are the special case where the interface is a monomial
`O X^ I`.

## Definitions

`DynSystem p` bundles the lens data with the state existential:

- `State : Type`,
- `expose : State → p.A` — the position presented at each state,
- `update : (s : State) → p.B (expose s) → State` — the next state given a
  direction at the exposed position. The dependence on `expose s` is what lets the
  available moves vary with the current position.

`toLens` / `ofLens` give the definitional round-trip with `selfMonomial State ⟹ p`.

- `MooreMachine O I = DynSystem (monomial O I)` — constant directions, so
  `expose : State → O` is the output and `update : State → I → State` is the
  transition.
- `DeterministicAutomaton O I` — a Moore machine with a start state; `accepts` is
  decidable language membership for `O = Bool`.
- `Closed = DynSystem X` — interface `X`, so an autonomous system: `step` is a
  pure endofunction and `iterate` its orbit.
- `Point p` (`X ⟹ p`) and `Section p` (`p ⟹ X`, via `sectionLens`) — generalized
  elements and dependent sections `(a : p.A) → p.B a` of an interface; a section
  is what closes a system.

## Combinators (§4.3–4.4)

Each is defined on the unpacked `expose` / `update` data and related to its lens
counterpart by a `*_toLens` lemma:

- `wrap` (§4.3.3) — re-interface along a lens `p ⟹ q` (postcomposition;
  `wrap_id` / `wrap_comp` are `rfl`).
- `close` / `MooreMachine.feedback` (§4.3.4) — close a system with a section; for
  a Moore machine the section is a feedback map `O → I`.
- `tensor` (§4.3.2) — parallel product; states multiply, interfaces tensor.
- `pairing` (§4.3.1) — categorical product; two interfaces on one shared state.
- `Wiring₂` / `wire₂` (§4.4) — a wiring diagram is a lens `p ⊗ q ⟹ r`; installing
  systems is `tensor` then `wrap`.

## Semantics

- Finite runs (`Run.lean`): `run`, `trace`, `outputOn`, `accepts`; input streams
  `stateStream` / `outputStream` shown to agree with finite `run` on every prefix
  (`stateStream_eq_run`).
- Infinite behaviour (`Trajectory.lean`): `trajectory` unfolds a system into the
  cofree-`p` tree `CofreeC p p.A`; for a closed system its spine is the state
  `iterate` (`next_iterate_trajectory`).
- Closed-loop Moore behaviour (`Behavior.lean`): `feedbackStream` and the
  identification of its trajectory spine with the iterated state.

`Examples.lean` exercises the API as regression tests: counter, parity automaton,
a mode-dependent (non-Moore) `gate` system, feedback / streams.

## Use in VCV-io

[VCV-io#460](https://github.com/Verified-zkEVM/VCV-io/pull/460) uses this layer
for the coalgebraic dual of its inductive `OracleComp`. An oracle signature is a
polynomial functor, so an `OracleStrategy` is a `PFunctor.DynSystem
spec.toPFunctor` and an oracle handler is a `PFunctor.Section`. Running a strategy
under a handler gives the `iterate = simulateQ` correspondence: a finite program
computes the same result read as inductive `OracleComp` syntax or as a dynamical
system's trajectory. That correspondence is what VCV-io builds oracle machines and
Turing-machine-grounded polytime adversaries on. Keeping the layer here (generic,
crypto-free) lets VCV-io depend on it without pulling probabilistic or security
content into PolyFun.

## Supporting changes

- Module docstring for `Control/Monad/FreeCont.lean`.
- `Control/Monad/Free.lean` imports `Mathlib.Tactic.GCongr` instead of all of
  `Mathlib.Tactic`.
- `scripts/lint-style.py` skips the module-system prologue (`module` keyword and
  `public`/`meta`/`private import` lines) in the header-layout check.
- `REFERENCES.md`, `docs/wiki/pfunctor.md`, `AGENTS.md` document the new layer.


dtumad:
@alexanderlhicks It would be nice if this could get merged into PolyFun before we bump for v4.31 on VCVio if we are already updating dependencies there.

alexanderlhicks (approved):

PR #17, opened 2026-07-11
feat(pfunctor): fuel-exact seqComp run law (ResolvesIn, runWith_seqComp)

dtumad:
Stacked on #16. This PR supplies the resolution-certificate algebra needed to turn PointedMachine sequencing into an exact bind principle.

- `ResolvesIn k st` means every answer path resolves within the query budget.
- `resolvesIn_iff_exists_toComp_eq_map_some` is now a true equivalence: resolution is exactly an all-`some` leaf factorization. The dependent continuation proof uses `FreeM.roll_inj`; classical choice is localized to extracting a proof-relevant tree family from the Prop-valued certificate.
- Resolution is monotone and composes additively through `seqComp`, including initial-state and second-phase lemmas.
- `runWith_eq_of_resolvesIn` gives fuel irrelevance after resolution.
- `runWith_seqComp_inl` / `runWith_seqComp_init` give the fuel-exact bind law at budget `k₁ + k₂` under the precise per-phase resolution hypotheses.
- Tests cover both leaf directions, exact/zero/surplus budgets, immediate and halted handoffs, lossy handlers, and universe elaboration.
- A concrete `StateT Nat Id` test runs two one-query phases and computes `(some 9, 2)`, proving that the sum is phase-local machine state while one ambient handler state is shared across the handoff.

Validated with focused builds, full delegated tests, docs/import checks, and style lint.


quangvdao:
## Follow-up design ledger

This pass deliberately folded the immediately canonical PointedMachine surface into this PR:

- input contravariance, output covariance, combined dimap, and interface transport along a Lens;
- simp lemmas for their observable fields;
- a closed deterministic resolution theorem identifying ResolvesIn with readable output along Closed.iterate;
- a Collatz example whose machine formulation is proved equivalent to the familiar reaches-one formulation.

The following are intentionally not part of this PR, but should remain visible for later work:

1. **Identity / pure machines.** There is no canonical generic pure for an arbitrary interface p: PointedMachine requires expose : State → p.A even in an already halted state, and p.A may be empty. A constructor parameterized by a chosen resting position may be useful, but should not be presented as a canonical monadic unit without an additional pointed-interface assumption or a semantic quotient that ignores exposure after output.

2. **Semantic Kleisli view.** Package execution from init as an OptionT/Kleisli-style program, then state and prove identity and associativity at the semantic level and connect seqComp to FreeM bind/append. The current PR supplies operational and fuel-exact laws, but the correct equality notion for machines needs an explicit decision.

3. **Termination vocabulary.** ResolvesIn is bounded must-resolution: every answer path resolves within the budget. May-resolution, divergence, and unbounded/coinductive execution should be introduced together with adequacy lemmas, not as isolated aliases.

4. **Stateful resource composition.** Shared global state such as a random-oracle table belongs below machine composition in a stateful handler. A future design should examine handler linking, resource frames, and shared-versus-private state directly instead of extending the Sum state used for phase-local control.

5. **Effectful dynamical systems.** A Kleisli/effectful generalization may unify environment actions, probabilistic responders, and stateful games, but it needs a named downstream consumer and a design comparison with the handler layer before choosing the abstraction boundary.

6. **Richer composition theory.** Cofree/comonoid structure, mates or retrofunctors, comodules, and bicomodule composition remain higher-risk theory work. Likewise, a terminal-state-dependent handoff should only be added if a concrete consumer shows that returning the handoff value or using ambient handler state is insufficient.

7. **Complexity.** Runtime and polynomial-time closure belong downstream in VCVio; PolyFun should provide the structural and semantic composition laws they consume.

Later stack PRs already cover adjacent pieces: #19 treats dynamical systems as lenses, #20 adds responder/game execution, #21 develops mapM universal properties, #22 adds seqComp state simp support, and #27 connects simulations/bisimulations to UC-style views. Those should be reviewed and restacked rather than duplicated here.

Issue #33, opened 2026-07-13
Make ITree maximally universe-polymorphic and complete its foundational API


quangvdao:
@dtumad This is a focused follow-up to #32. That issue needs a universe-polymorphic `Resumption` and an honest bridge into `ITree`; while checking that boundary, it became clear that the present `ITree` API collapses several independent universes and also stops short of the semantic API we will need to use it as foundational plumbing.

This issue is anchored to `main` at [`f1a4c4c2e8bbd3308cecc869a20f6ddefaff4e38`](https://github.com/Verified-zkEVM/PolyFun/commit/f1a4c4c2e8bbd3308cecc869a20f6ddefaff4e38), the state described by #32.

The primary goal is **to make `ITree` maximally universe-polymorphic, definition by definition rather than by a blanket same-universe convention**. Since this cutover touches the entire ITree stack, the issue also records the important foundational API gaps that should be closed while or immediately after we generalize it.

## Current restriction

Every current ITree module starts from essentially one universe:

```lean
F : PFunctor.{u, u}
R : Type u
ITree F R : Type u
```

In particular:

```lean
inductive ITree.Shape (F : PFunctor.{u, u}) (R : Type u) : Type u
def ITree.Poly (F : PFunctor.{u, u}) (R : Type u) : PFunctor.{u, u}
def ITree (F : PFunctor.{u, u}) (R : Type u) : Type u
```

The same alignment is repeated through `Construct`, `Handler`, `Sim`, `Bisim`, `Rec`, `Events`, and `Unfold`. The wiki currently presents this as part of the model, but it is not inherent to interaction trees or to `PFunctor.M`.

There are three independent pieces of data:

- event names / polynomial positions `F.A : Type uA`;
- replies / polynomial directions `F.B a : Type uB`;
- returned values `R : Type uR`.

Conflating them prevents ordinary examples such as a small command type whose reply carries a large object, a large recursive-call input returning a small result, or an exception type in a different universe from the computation's result.

## The maximally polymorphic core

I checked the following definitions in a standalone Lean file against the current dependencies. They elaborate and compile:

```lean
universe uA uB uR

inductive ITree.Shape (F : PFunctor.{uA, uB}) (R : Type uR) :
    Type (max uA uR) where
  | pure (r : R)
  | step
  | query (a : F.A)

@[reducible]
def ITree.Poly (F : PFunctor.{uA, uB}) (R : Type uR) :
    PFunctor.{max uA uR, uB} where
[...]

Issue #32, opened 2026-07-13
Design: returning dynamical computations, resumptions, and their FreeM/ITree semantics


quangvdao:
@dtumad This issue records a design tension exposed while reviewing the new dynamical-system stack and proposes a canonical separation between stateful realizations, state-free behavior, finite programs, and interaction trees.

This is deliberately a design issue rather than an implementation PR: the central proposal seems strong, but the names, universe-polymorphic surface, and exact migration boundaries should be reviewed before cutting over downstream users.

## Repository state this issue describes

This issue is anchored to `main` at commit [`f1a4c4c2e8bbd3308cecc869a20f6ddefaff4e38`](https://github.com/Verified-zkEVM/PolyFun/commit/f1a4c4c2e8bbd3308cecc869a20f6ddefaff4e38), the squash merge of #31 (“adopt upstream cslib free monad definitions”), on 2026-07-13.

At that commit:

- `PFunctor.FreeM` is the upstream cslib inductive free monad, with PolyFun retaining its polynomial-specific maps, handlers, paths, displayed families, and roll bounds.
- `PFunctor.DynSystem S p` is definitionally `Lens (selfMonomial S) p`, equivalently the coalgebra `S → p.Obj S`.
- `DynSystem.behavior : S → PFunctor.M p` gives the unique terminal-coalgebra behavior tree.
- `ITree p β` is the M-type with `Ret`, `Tau`, and `Vis` nodes.
- `DynSystem.toITree` currently embeds a non-returning dynamical system as an all-query `ITree p PEmpty`.
- `PointedMachine p α β` still stores `State`, `expose`, `update`, `init`, and a separate `output : State → Option β`.
- #29 is the only remaining open PR. It adds `PointedMachine.Implements` and its simulation proof method, but is still phrased against the current representation and the pre-#31 branch tip. Its proof ideas should be preserved and restacked after this design is settled.

No existing open issue covers this proposal.

## The current tension

The current definition is:

```lean
structure PointedMachine (p : PFunctor) (α β : Type) where
  State  : Type
  expose : State → p.A
  update : (s : State) → p.B (expose s) → State
  init   : α → State
  output : State → Option β
```

Operationally, a state with `output s = some b` has returned and the runner stops. Structurally, however, that state must still expose a `p`-position and provide updates for all its directions.

This causes several concrete problems.

### 1. Returned states contain unreachable interface data

An already-returned computation still needs `expose` and `update`. The clearest symptom is:

```lean
[...]

PR #34, opened 2026-07-13
refactor(dynamical): make Machine an extensible base substrate

dtumad:
## Summary

This refactors `Machine` into the minimal bundled substrate for dynamical systems:

```lean
structure Machine (p : PFunctor) where
  State : Type u
  behavior : DynSystem State p
```

The base bundle deliberately contains only operational dynamics. More specialized notions now extend it with the additional structure they need:

- `IOMachine p α β extends Machine p` adds `init` and partial `output`.
- `Labeled p extends Machine p` adds stable transition events.
- `Ticketed p extends Machine p` adds scheduling-obligation identifiers.
- `SafetySpec p extends Machine p` adds initial-state, assumption, and safety predicates.

This makes `Machine` a reusable substrate instead of identifying all machines with the input/output use case. Consumers can define new extensions without inheriting irrelevant initialization or readout fields, while all bundles expose their underlying dynamics uniformly through `.toMachine.behavior`.

## Refactor details

- Renames the previous `PointedMachine` module and structure to `IOMachine`.
- Removes bundle-specific `toDynSystem`, `expose`, and `update` forwarding APIs in favor of the common `.toMachine.behavior` path.
- Migrates safety refinement and concurrent process adapters to the shared hierarchy.
- Uses record updates to preserve the parent bundle when changing only an extension field or wrapped behavior.
- Expands the cleaner `⇆` lens literal and diagrammatic `⨟` composition notation across the dynamical implementation and examples.
- Updates generated imports, worked examples, the repository guide, and long-form documentation for the new hierarchy.

## Why this shape

`State` and `behavior` are the common denominator across dynamical consumers. Initialization, output, labels, tickets, and verification predicates are independent decorations on that core. Encoding that distinction in the structure hierarchy keeps the operational substrate small and lets each layer state exactly which additional capabilities it requires.

The uniform projection also makes generic code straightforward: regardless of the extension in use, the underlying system is always `M.toMachine.behavior`.

## Validation

- `lake build PolyFunTest --wfail`
- `lake exe lint-style`
- `git diff --check`


dtumad:
@quangvdao I think this at least sets up a reasonable base naming system before #32 and #33 invest in more heavy modification. Just meant to seperate out a more minimal stack of gadgets so that minimal shared components can hopefully be shared more easily.

PR #44, opened 2026-07-15
Generalize ITree handlers and simulation

quangvdao:
## Summary

Based on merged #42 and current `main`. Review only the unique diff on `agent/itree-handler-sim`.

- generalize `ITree.Handler`, `simulate`, and `mapSpec` across independent source, target, and result universes
- add coproduct handler routing with `Handler.case_`
- add handler composition with `Handler.comp` and its definitional distribution over routing
- generalize the primitive strong simulation/mapSpec equations and add pointwise identity and pure-renaming composition laws
- add universe compile canaries plus observable routing/composition producer tests
- update the ITree guide and correct its descriptions of the existing State/Exception event modules

## Why

This is I2 of the interaction-tree universe/generalization sequence from #33. It moves the executable handler/simulation layer off the homogeneous `PFunctor.{u, u}` restriction while adding the routing and composition APIs needed by later effect-handler work.

## Boundary

The existing `WeakBisim` relation and its bind/iteration theory are still homogeneous in this slice. Consequently, laws that consume that algebra—including `simulate_query_eq_bind`, `simulate_query`, `mapSpec_bind`, `mapSpec_iter`, `mapSpec_map`, and `mapSpec_cat`—remain homogeneous here. #45 generalizes the underlying bind/iteration and relational bisimulation layer; #46 then generalizes these simulation/mapSpec consumer laws and adds separated-universe canaries. General handler associativity and `simulate_comp` also remain in #46.

## Validation

- `lake exe lint-style`
- `lake env lean PolyFunTest/ITree/HandlerSimulationExamples.lean`
- targeted build of `PolyFun.ITree.Sim.Facts`, State/Exception events, and `PolyFun.ITree.Rec`
- `git diff --check`

Part of #33.

### DIFF
--- a/PolyFun/ITree/Bisim/Bind.lean
+++ b/PolyFun/ITree/Bisim/Bind.lean
@@ -10,13 +10,18 @@ public import PolyFun.ITree.Bisim.Equiv
 /-! # Algebraic laws for `bind` and `iter`
 
 The classical equational theory of interaction trees, lifted to Lean. All
-laws are stated either as strong bisimulations (`Bisim`, i.e. definitional
+laws are stated either as strong bisimulations (`Bisim`, i.e. Lean
 equality of M-types) or weak bisimulations (`WeakBisim`).
+
+The event-position and event-direction universes and every result type used by
+the named `bind`/`iter` laws are independent. `bind_weakBisimRel` exposes the
+fully relational theorem; `bind_weakBisim_cont` is its equality-specialized,
+one-sided corollary.
 
 ## Main statements
 
 * `bind_pure_left`, `bind_pure_right`, `bind_assoc` — monad laws on
-  `ITree.bind`, as strong bisimulations (i.e. definitional equalities on
+  `ITree.bind`, as strong bisimulations (i.e. exact equalities on
   `PFunctor.M`).
 * `bind_step`, `bind_query` — `bind` distributes over a leading silent
   step / visible query.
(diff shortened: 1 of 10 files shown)


dtumad:
There are a few additional universe-polymorphism gaps beyond the `simulate_query_eq_bind` / `simulate_query` case already noted in the existing review thread.

The non-notation versions of laws `mapSpec_bind`, `mapSpec_iter`, `mapSpec_map`, and `mapSpec_cat` still inherit the module-level `{E F G : PFunctor.{u, u}} {α β : Type u}` variables. As a result, clients can use the newly generalized `mapSpec` and the mixed-universe named `bind`, `iter`, `map`, and `cat` operations, but cannot apply their corresponding compatibility laws when the signature or value universes differ.

This kinda goes back to the `>>=` vs. `.bind` issue as well.

quangvdao:
Addressed in the rebuilt stack. These consumer laws require the named bind/iteration algebra that is generalized in #45, so #44 now documents that dependency boundary explicitly. #46 (`8e3e8f8`) generalizes `simulate_query_eq_bind`, `simulate_query`, `mapSpec_bind`, `mapSpec_iter`, `mapSpec_map`, and `mapSpec_cat`, and adds separated-universe producer canaries for query, bind, map, iter, and cat. This keeps #45 reviewable as the foundational relational/bind layer while closing the API gaps before downstream recursion consumers.

PR #54, opened 2026-07-15
test(pfunctor): harden free polynomial path laws

quangvdao:
## Why this PR changed

The original free-polynomial construction was absorbed into `main` through
#64, including the substitution-monoid structure and its universal-property
foundation. Keeping the old branch would therefore duplicate already-landed
code and conflict with `main`.

This PR has been rebuilt on current `main` as the remaining review-hardening
slice.

## What remains in this PR

- Keep `FreeP.map_comp` out of the simp set so `Lens.mapObj_comp` determines
  the canonical object-level normal form.
- Add a direct `FreeP.map.toFunB` canary for a branch-reversing generator
  lens.
- Test multiplication path splitting with distinct Boolean outer and inner
  directions; the previous unary test could not detect swapping or copying.
- Test both sides of three-layer associativity on three independently
  observable backward path components.
- Record the actual `Lens.Basic -> SubstMonoid` and
  `{Free.Path, SubstMonoid} -> Free.Polynomial` import edges in both module-DAG
  documents.

## Review outcome

The landed construction correctly presents finite `P`-trees as positions,
complete paths as directions, and grafting/path splitting as substitution
multiplication. The full lens unit and associativity laws cover both forward
tree behavior and dependent backward path behavior. No mathematical or trust
boundary blocker was found.

This hardening delta specifically closes the remaining executable-coverage,
simp-normalization, and documentation issues found during review.

## Validation

- `lake exe lint-style`
- direct compilation of `PolyFunTest/PFunctor/FreePolynomial.lean`
- `git diff --check`
- no added `sorry`, `admit`, `unsafe`, or axioms

## Stack

This remains the oldest open PR in the pattern-runs-on-matter train. #55 is
the next logical slice and will be compared against current `main` and
restacked after this PR merges.

### DIFF
--- a/PolyFun/PFunctor/Free/Polynomial.lean
+++ b/PolyFun/PFunctor/Free/Polynomial.lean
@@ -333,7 +333,9 @@ theorem map_obj_comp (l₂ : Lens Q R) (l₁ : Lens P Q)
         (map_obj_comp l₂ l₁
           (rest (l₁.toFunB a (l₂.toFunB (l₁.toFunA a) d))))
 
-@[simp]
+/-- Mapping preserves lens composition. This is intentionally not a simp
+lemma: `Lens.mapObj_comp` already selects the canonical normal form for
+composition acting on polynomial extensions. -/
 theorem map_comp (l₂ : Lens Q R) (l₁ : Lens P Q) :
     map l₂ ∘ₗ map l₁ = map (l₂ ∘ₗ l₁) := by
   let hA : ∀ s, (map l₂ ∘ₗ map l₁).toFunA s =
--- a/PolyFunTest/PFunctor/FreePolynomial.lean
+++ b/PolyFunTest/PFunctor/FreePolynomial.lean
@@ -25,6 +25,18 @@ namespace FreePolynomialTest
 /-- A binary-branching signature whose operation positions are also bits. -/
 abbrev binaryP : PFunctor := ⟨Bool, fun _ => Bool⟩
 
+def binaryLeaf : FreeM binaryP PUnit :=
+  .pure PUnit.unit
+
+def binaryOp (label : Bool) : FreeM binaryP PUnit :=
+  .liftBind label fun _ => binaryLeaf
+
+/-- Three binary substitution layers with independently observable path
+components. -/
+def binaryTriple : ((FreeP binaryP ◃ FreeP binaryP) ◃ FreeP binaryP).A :=
+  ⟨⟨binaryOp false, fun _ => binaryOp true⟩,
+    fun _ => binaryOp false⟩
+
 /-- A binary tree with distinguishable payloads at its two leaves. -/
 def labelledBinaryTree : FreeM binaryP Nat :=
   .liftBind false fun branch => .pure (if branch then 7 else 3)
@@ -78,6 +90,31 @@ example :
     ⟨⟨PUnit.unit, ⟨⟩⟩, ⟨PUnit.unit, ⟨⟩⟩⟩ :=
   rfl
 
+/-- Splitting must preserve both nontrivial branch choices; a unary signature
+would not detect swapping or duplicating the outer and inner components. -/
+example :
+    (FreeP.mult (P := binaryP)).toFunB
+      ⟨binaryOp false, fun _ => binaryOp true⟩
+      ⟨false, true, ⟨⟩⟩ =
+    ⟨⟨false, ⟨⟩⟩, ⟨true, ⟨⟩⟩⟩ :=
+  rfl
+
+/-- Left-associated multiplication reassociates all three path components
+without permuting them. -/
+example :
+    (FreeP.multAssocLeft (P := binaryP)).toFunB binaryTriple
+      ⟨false, true, false, ⟨⟩⟩ =
+    ⟨⟨⟨false, ⟨⟩⟩, ⟨true, ⟨⟩⟩⟩, ⟨false, ⟨⟩⟩⟩ :=
+  rfl
+
+/-- The right-associated composite has the same observable reassociation of
+three nontrivial path components. -/
+example :
+    (FreeP.multAssocRight (P := binaryP)).toFunB binaryTriple
+      ⟨false, true, false, ⟨⟩⟩ =
+    ⟨⟨⟨false, ⟨⟩⟩, ⟨true, ⟨⟩⟩⟩, ⟨false, ⟨⟩⟩⟩ :=
+  rfl
+
 /-- A target signature with natural-number operation labels. -/
 abbrev natBinaryP : PFunctor := ⟨Nat, fun _ => Bool⟩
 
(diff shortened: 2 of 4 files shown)


PolyFun/PFunctor/Free/Polynomial.lean
```diff
@@ -0,0 +1,509 @@
...
+          fun pair => inner
+            ⟨FreeM.Path.cons a rest b pair.1, pair.2⟩⟩
+      let embed (b : P.B a) :
+          ((FreeP P ◃ FreeP P) ◃ FreeP P).B (child b) →
+          ((FreeP P ◃ FreeP P) ◃ FreeP P).B
+            ⟨⟨FreeM.liftBind a rest, middle⟩, inner⟩ :=
+        fun direction =>
+          ⟨⟨FreeM.Path.cons a rest b direction.1.1,
+            direction.1.2⟩, direction.2⟩
+      change
```
dtumad: It seems a lot of these proofs need `change` statements in their core, seems like maybe a smell for missing lemmas or `simp` paths?

PolyFun/PFunctor/Free/Polynomial.lean
```diff
@@ -0,0 +1,509 @@
...
+  match s with
+  | .pure u =>
+      cases u
+      refine Sigma.ext (by rfl) ?_
+      apply heq_of_eq
+      funext path
+      cases path
+      rfl
+  | .liftBind a rest =>
+      change
```
dtumad: Same here

PolyFun/PFunctor/Free/Polynomial.lean
```diff
@@ -0,0 +1,509 @@
...
```
dtumad: and here

PolyFun/PFunctor/Free/Polynomial.lean
dtumad: Resolved in b58228e by deriving right unitality from the decode/multiplication compatibility lemma and the FreeM monad laws; the pointwise object theorem is now only a short corollary.

PolyFun/PFunctor/Free/Polynomial.lean
dtumad: Resolved in b58228e. I added the reusable Lens.mapObj/ext_mapObj API and refactored the FreeP laws through decode, FreeM.mapLens, and monadic join/associativity. Polynomial.lean no longer contains any change proof blocks.

PolyFun/PFunctor/Free/Polynomial.lean
dtumad: Resolved in b58228e. Associativity now decodes both polynomial composites to the two sides of FreeM.bind_assoc, with the pointwise object theorem retained as a corollary.

PolyFunTest/PFunctor/FreePolynomial.lean
```diff
@@ -0,0 +1,150 @@
...
+def reverseBranchLens : Lens binaryP natBinaryP :=
+  (fun label => if label then 1 else 0) ⇆ (fun _ branch => !branch)
+
+def controlTree : FreeM binaryP PUnit :=
+  .liftBind false fun branch =>
+    .liftBind branch fun _ => .pure PUnit.unit
+
+/-- The backward lens direction is observable: runtime `false` selects the
+source `true` child (label `1`), while runtime `true` selects label `0`. -/
+example : FreeP.mapShape reverseBranchLens controlTree =
```
dtumad, line 132: The file's stated purpose includes the backward direction of a nontrivial generator lens, and `mult.toFunB` does get a direct test above — but `(FreeP.map reverseBranchLens).toFunB` is never evaluated directly; this example observes only the forward `mapShape` (the backward map is exercised indirectly via child selection). A direct `example : (FreeP.map reverseBranchLens).toFunB controlTree ⟨…⟩ = ⟨…⟩ := rfl` would pin the path-pullback component too.

dtumad (commented):
Follow-up review of the full diff after b58228e. The earlier `change`-smell feedback is genuinely resolved: `Free/Polynomial.lean` now contains zero `change`/`cast`/`Eq.mpr`/`▸`, associativity and right unit route through `decode_mult`/`decode_relabel` into `FreeM.bind_assoc`/`bind_pure_comp`, and the dependent-type friction is concentrated once in the new `Lens.ext_mapObj` bridge (the remaining `change`s in Path.lean pre-exist on main). Left unit is `rfl`. The monoid laws are full lens equalities covering both position and direction maps, and `FreeP.map`'s three universe pairs are genuinely independent and pinned by tests. One inline item; three more here:

1. **`Path.lean:419` `append_tree_assoc`**: lives in `namespace Path` but its head symbol is `FreeM.append`, so the name reads as a lemma about `Path.append` against the `{head_symbol}_{operation}` convention — and nothing in this PR consumes it besides its own recursive call. If it's staged for the stacked PR (#66 does use it), consider moving it out of `namespace Path` as `FreeM.append_assoc`.

2. **Simp orientation**: `@[simp] Lens.mapObj_comp : mapObj (l₂ ∘ₗ l₁) x = mapObj l₂ (mapObj l₁ x)` together with `@[simp] FreeP.map_comp : map l₂ ∘ₗ map l₁ = map (l₂ ∘ₗ l₁)` makes the simp set non-confluent on `Lens.mapObj (FreeP.map l₂ ∘ₗ FreeP.map l₁) x`. Worth settling the orientation before downstream files bake in a normal form.

3. **AGENTS.md Module Layering**: the diagram gains no entry for `Free/Polynomial` (and #53 never added `SubstMonoid`). Suggest `PFunctor/{Lens/Basic, Free/Path} → PFunctor/SubstMonoid → PFunctor/Free/Polynomial`; the same sync gap exists on several PRs in this stack, so one pass covering the whole stack may be easiest.

4. **Draft gate**: the PR body still says "intentionally left as a draft and must not be merged while the full stack is being assembled", but the PR is marked ready for review — worth updating the body so the merge gate is unambiguous.

PolyFunTest/PFunctor/FreePolynomial.lean
quangvdao: Addressed in `7fb9e6f`. I added a direct `rfl` canary for `(FreeP.map reverseBranchLens).toFunB` on `controlTree`: the target path `⟨false, true, ⟨⟩⟩` pulls back to `⟨true, false, ⟨⟩⟩`. This exercises the backward path map itself, independently of `mapShape`.

quangvdao:
Response to @dtumad's follow-up review (current head `7fb9e6f`):

1. Moved the grafting theorem out of `namespace Path` and renamed it `FreeM.append_assoc` in `7fb9e6f`. I intentionally did not retain a compatibility alias for the unmerged mis-owned name. The separate #66 branch still has two old-name call sites to update when that branch is restacked.
2. Removed `[simp]` from `FreeP.map_comp` and documented that `Lens.mapObj_comp` selects the canonical nested-object normal form, eliminating the non-confluent simp pair.
3. Added the actual import edges to `AGENTS.md`: `Lens/Composite → SubstMonoid` and `{Free/Path, SubstMonoid} → Free/Polynomial`.
4. Kept the PR draft while the train was being assembled. The full stack is now complete, the user explicitly requested all PRs ready, and I updated the PR body accordingly.

I replied separately on the inline backward-map thread with the exact new direct `toFunB` canary.

PR #55, opened 2026-07-15
feat(pfunctor): sharpen free polynomial universal property

quangvdao:
## Why this PR changed

The original free-polynomial universal-property implementation was absorbed
into `main` through #64. Its old stacked history therefore duplicated landed
code and conflicted with the current repository.

This PR has been rebuilt on current `main` as the remaining API and proof
hardening from review.

## What remains in this PR

- Add `Lens.ext_mapObj`, a reusable dependent extensionality theorem for
  proving lens equality from equality on identity-labelled polynomial
  objects.
- Route four universal-property proofs through that abstraction, removing 80
  lines of repeated `HEq` and transport plumbing.
- Add `FreeP.extend_unique`: agreement on the one-node generators uniquely
  determines a homomorphism out of the free substitution monoid.
- Add a direct Boolean canary for `(FreeP.extend ...).toLens.toFunB` on a
  two-node tree, making path splitting and both branch reversals observable.

## Review outcome

The landed construction proves the genuine free-object property. `extend`
is a substitution-monoid homomorphism, `restrict` recovers its action on
generators, and the two inverse laws are established as full lens equalities,
including dependent backward direction maps. The uniqueness direction is
routed through `FreeM.liftM_natural`, which is the correct abstraction layer.

The existing homogeneous `SubstMonoid.Hom` universe boundary is preserved;
this PR makes no claim to generalize that upstream interface. No mathematical
or trust-boundary blocker was found.

## Validation

- `lake exe lint-style`
- targeted builds of `Lens.Basic`, `Free.Universal`, and its producer tests
- `git diff --check`
- no added `sorry`, `admit`, `unsafe`, or axioms
- principal axiom footprints remain `propext`, `Classical.choice`, and
  `Quot.sound`

## Stack

This is now the oldest open PR in the pattern-runs-on-matter train. #56 is the
next semantic slice and develops finite vertices of polynomial M-types.

### DIFF
--- a/PolyFunTest/PFunctor/FreeUniversal.lean
+++ b/PolyFunTest/PFunctor/FreeUniversal.lean
@@ -64,6 +64,12 @@ example (f : SubstMonoid.Hom (FreeP.substMonoid bitUnary) wordMonoid) :
     FreeP.extend wordMonoid (FreeP.restrict wordMonoid f) = f :=
   FreeP.extend_restrict wordMonoid f
 
+/-- Agreement on generators determines a homomorphism out of `FreeP`. -/
+example (f : SubstMonoid.Hom (FreeP.substMonoid bitUnary) wordMonoid)
+    (h : f.toLens ∘ₗ FreeP.generator bitUnary = bitToWord) :
+    f = FreeP.extend wordMonoid bitToWord :=
+  FreeP.extend_unique wordMonoid bitToWord f h
+
 /-- The packaged equivalence exposes both inverse laws. -/
 example : (FreeP.homEquiv wordMonoid).symm
     ((FreeP.homEquiv wordMonoid) (FreeP.extend wordMonoid bitToWord)) =
@@ -83,6 +89,15 @@ def oneBinaryNode : (FreeP binaryP).A :=
 the nontrivial backward component, not only the mapped shape. -/
 example : (FreeP.mapHom reverseBranchLens).toLens.toFunB oneBinaryNode
     ⟨false, ⟨⟩⟩ = ⟨true, ⟨⟩⟩ :=
+  rfl
+
+/-- The universal fold's backward interpreter splits a two-node target path
+through free multiplication and reverses both target branches. This pins
+`foldPath` directly, rather than restating `FreeP.map.toFunB`. -/
+example :
+    (FreeP.extend (FreeP.substMonoid natBinaryP)
+      (FreeP.generator natBinaryP ∘ₗ reverseBranchLens)).toLens.toFunB
+        controlTree ⟨false, true, ⟨⟩⟩ = ⟨true, false, ⟨⟩⟩ :=
   rfl
 
 /-- The polynomial fold is definitionally the existing `FreeM.liftM` fold on
--- a/PolyFun/PFunctor/Free/Universal.lean
+++ b/PolyFun/PFunctor/Free/Universal.lean
@@ -205,29 +205,9 @@ including its backward path map. -/
 theorem foldLens_comp_mult :
     foldLens M l ∘ₗ mult (P := P) =
       M.mult ∘ₗ (foldLens M l ◃ₗ foldLens M l) := by
-  let hA : ∀ x : (FreeP P ◃ FreeP P).A,
-      (foldLens M l ∘ₗ mult (P := P)).toFunA x =
-        (M.mult ∘ₗ (foldLens M l ◃ₗ foldLens M l)).toFunA x :=
-    fun x => by
-      rcases x with ⟨s, next⟩
-      exact congrArg Sigma.fst (foldLens_mult_obj M l s next)
-  refine Lens.ext _ _ hA ?_
-  intro x
-  rcases x with ⟨s, next⟩
-  apply eq_of_heq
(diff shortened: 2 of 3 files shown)


dtumad (commented):
Reviewed the full diff plus the rebase situation. The mathematics is sound: `extend`/`restrict` give the genuine free-object bijection with uniqueness proved for an arbitrary hom (`extend_restrict`), the backward-path law is genuinely dependent, and routing uniqueness through `FreeM.liftM_natural` rather than a fresh induction is elegant. `Extension.lean`'s every-SubstMonoid-is-a-LawfulMonad is independently valuable and notably cast-free.

**Rebase roadmap** (base moved by b58228e + a main merge; merge-base `3461b850`):
- The only *textual* conflict is `docs/wiki/pfunctor.md` (both branches rewrote the `Lens/Basic.lean` table row; after merging, `Free/Polynomial.lean` would also be documented in two tables — dedupe to one, the new SubstMonoid table is arguably the better home).
- **Trap**: `PolyFun/PFunctor/Lens/Basic.lean` merges *cleanly* and then fails to build — b58228e added an identical `Lens.mapObj`/`mapObj_id`/`mapObj_comp` block at a different location (base lines 82–100, which additionally has `ext_mapObj`). Drop this PR's hunk wholesale; base's version is a superset.
- `decodeAt_bind`/`decodeAt_append_split` and the new files use only names untouched by b58228e, so no structural proof rework is needed.
- **Style pass**: `Universal.lean` predates b58228e and contains four verbatim copies of the HEq pattern (`Lens.ext` + `eq_of_heq`/`Sigma.ext_iff`/`eqRec_heq_self`) that b58228e extracted into `Lens.ext_mapObj` — at lines 214–230, 262–270, 349–363, 469–480. Each collapses to roughly `Lens.ext_mapObj _ _ (fun ⟨s, next⟩ => foldLens_mult_obj M l s next)` since the pointwise lemmas are already stated in `mapObj` form. Shipping without this would just re-open the same review thread that b58228e closed on #54.

Two smaller API notes: a named `extend_unique : f.toLens ∘ₗ generator P = l → f = extend M l` (two lines from `extend_restrict`) and/or a `hom_ext` would match the `FreeM.liftMHom_unique` precedent one layer down; and `homEquiv`'s forward direction is `restrict` (Hom → Lens), the reverse of Mathlib's `FreeGroup.lift`-style convention — fine, but worth a conscious choice.

The homogeneous-`SubstMonoid.Hom` universe boundary (which #65/#67/#69/#70 keep re-documenting) deserves a single upstream decision — raising that in the stack-level review with @dtumad rather than here.

PolyFunTest/PFunctor/FreeUniversal.lean
```diff
@@ -0,0 +1,104 @@
...
+
+/-- The universal fold respects outer-before-inner substitution order. -/
+example : (FreeP.extend wordMonoid bitToWord).toLens.toFunA twoBits =
+    [false, true] :=
+  rfl
+
+/-- Bind in the extension monad has the same observable order. -/
+example :
+    let outer : SubstMonoid.Extension wordMonoid Bool :=
+      ⟨[false], fun _ => true⟩
```
dtumad, line 50: `wordP` has `fun _ => PUnit` directions, so these orientation/bind-order tests can't falsify a backward-map bug, and the one backward test below exercises `FreeP.map.toFunB` (established in #54; `mapHom.toLens := map lPQ` is `rfl`) rather than the fold's `foldPath` against a nontrivial `M.mult.toFunB`. A `rfl`-checkable test of `(FreeP.extend … ).toLens.toFunB` on a two-node `binaryP` tree would make a split-order or branch mistake observable — `binaryP`'s `Bool` directions are enough.

quangvdao: Addressed in `279a5b5` (current PR head `2de35e6`). `FreeUniversal.lean` now directly evaluates `(FreeP.extend …).toLens.toFunB` on the two-node Boolean `controlTree`; both nontrivial branch choices reverse, so split order or a faulty `foldPath` is observable.

quangvdao:
Response to @dtumad's review (current head `2de35e6`):

- Restacked onto the updated #54 without duplicating the new `Lens.mapObj` block; the base's superset, including `Lens.ext_mapObj`, is the only retained definition.
- Deduplicated the `Free/Polynomial.lean` wiki entry.
- Replaced all four copied HEq/lens-extension proof blocks in `Free/Universal.lean` with `Lens.ext_mapObj`.
- Added `FreeP.extend_unique : f.toLens ∘ₗ generator P = l → f = extend M l` and a producer canary. I did not add a redundant `hom_ext`: `SubstMonoid.Hom.ext` already provides hom extensionality, and the review requested `extend_unique` and/or `hom_ext`.
- Kept `homEquiv` oriented Hom → Lens deliberately: its forward map is explicitly named `restrict`, its inverse is `extend`, and both inverse laws are tested.
- Centralized the homogeneous `SubstMonoid.Hom` universe restriction in the structure's upstream documentation instead of repeating it downstream.

The two-node nontrivial `FreeP.extend.toLens.toFunB` regression is described in the inline reply.

PR #58, opened 2026-07-16
feat(displayed): restrict displayed data along cursors

quangvdao:
## What changed

- add `Displayed.Shape.ChildProjection` and dependent `Displayed.OverShape.ChildProjection`
- define generic cursor-spine restriction once against those capabilities
- specialize restriction to `Decoration` and `Decoration.Over`
- prove root/down, composition, map/base-map, and `ofOver`/`toOver` compatibility laws
- add dependent cursor-restriction examples and repository documentation

## Why

An arbitrary displayed shape cannot be restricted along a cursor: a value of `D.node a child` need not expose the selected `child b`. Making child projection an explicit capability preserves the right abstraction boundary while allowing canonical decoration instances.

## Impact

This supplies PF-2 for downstream reachable-prefix work. Restriction returns data on the selected future residual subtree; visited-prefix history remains a separate fold over the cursor spine.

## Validation

- `lake build PolyFunTest.PFunctor.FreeDisplayedCursorExamples`
- full `lake build` (8,813 jobs)


docs/wiki/repo-map.md
```diff
@@ -45,6 +45,7 @@ PFunctor/Free/Basic
   -> PFunctor/Free/{Path, Displayed/Decoration}
   -> PFunctor/Free/Path/Execution
   -> PFunctor/Free/Cursor
+  -> PFunctor/Free/Displayed/Cursor
```
dtumad, line 48: This adds `PFunctor/Free/Displayed/Cursor` to repo-map.md's copy of the import DAG, but the mirrored Module Layering block in `AGENTS.md` doesn't get the same entry (`grep Displayed/Cursor AGENTS.md` has no hit on this branch). The two blocks are line-for-line mirrors on main, and the wiki contract asks for them to be updated in the same PR — could you sync the entry into AGENTS.md as well? (Optionally as a branch line `PFunctor/Free/{Cursor, Displayed/Decoration} -> PFunctor/Free/Displayed/Cursor`, since the linear insertion implies `Cursor/Occurrence` sits downstream of this file, which isn't a real edge.)

PolyFunTest/PFunctor/FreeDisplayedCursorExamples.lean
```diff
@@ -0,0 +1,96 @@
...
+-/
+module
+
+public import PolyFun.PFunctor.Free.Displayed.Cursor
+
+/-! # Cursor restriction examples for displayed free programs -/
+
+@[expose] public section
+
+universe uA uB v
```
dtumad, line 14: `universe uA uB v` is declared but never used — all examples in this file are at `Type 0` via `Command`/`Nat`. Either drop the line, or (nicer) follow the `FreeCursorExamples.lean` pattern and add an explicit universe-independence example, e.g. `Decoration.restrict` at fully general `PFunctor.{uA,uB}` with `α : Type v`, which would also exercise the generality claim.

dtumad (commented):
Reviewed the diff (capability record, generic restriction, Decoration specializations, laws, tests). Overall this looks high quality — the `ChildProjection` capability is the right abstraction boundary, the laws are stated at the correct dependency level, and the proofs are clean structural inductions. Two small items inline; some design questions (e.g. whether the `HEq` comp laws should be restated as `Sigma`/`Total`-space equalities) are being collected separately rather than blocking this PR.

PR #59, opened 2026-07-16
feat(free): classify cursors through append

quangvdao:
## What changed

- classify cursors through dependent `FreeM.append` into internal-prefix and selected-suffix cases
- provide cast-free `split`/`join` inverse laws, cursor composition, path, trace, length, and plug compatibility
- place `IsNode` and decoration restriction laws at their owning abstraction layers
- generalize displayed append across independent leaf universes
- add direct branch-sensitive cursor and path canaries
- synchronize the module DAG documentation

## Validation

- `lake exe lint-style`
- targeted cursor-append source and test builds
- diff/trust scan and principal `#print axioms` checks


PolyFun/PFunctor/Free/Cursor/Append.lean
```diff
@@ -0,0 +1,457 @@
...
+    split program suffix
+      (Cursor.ofPath (FreeM.append program suffix)
+        (Path.append program suffix path suffixPath)) =
+      .right path (Cursor.ofPath (suffix path) suffixPath) := by
+  rw [← joinRight_ofPath]
+  exact split_joinRight program suffix path _
+
+/-! ## Decoration compatibility -/
+
+namespace Decoration
```
dtumad, line 474: These restriction lemmas have head symbol `Displayed.Decoration.restrict`, whose sibling laws from #58 (`restrict_comp`, `restrict_map`, `restrict_ofOver`) live in `PFunctor.FreeM.Displayed.Decoration` — but this `namespace Decoration` is nested inside `Cursor`, so they land in the parallel `PFunctor.FreeM.Cursor.Decoration` (the test file ends up writing the fully-qualified `Cursor.Decoration.restrict_liftAppend` while using `Decoration.restrict` unqualified nearby). Mathlib convention puts lemmas where the head symbol lives; declaring these inside `namespace Displayed.Decoration` (and `.Over`) works from this file without changing the import DAG.

PolyFun/PFunctor/Free/Cursor/Append.lean
```diff
@@ -0,0 +1,457 @@
...
+universe uA uB v w w₂ w₃
+
+namespace PFunctor
+namespace FreeM
+namespace Cursor
+
+variable {P : PFunctor.{uA, uB}} {α : Type v} {β : Type w}
+
+/-- Evidence that a free program is an internal node rather than a leaf. -/
+inductive IsNode : FreeM P α → Prop
```
dtumad, line 34: `IsNode` is a predicate on `FreeM P α` and mentions no cursor, but is scoped under `Cursor`. Two suggestions: (a) move it up to the `FreeM` namespace; (b) add the bridge to the existing `Cursor.IsTerminal` (base branch `Free/Cursor.lean:473`), e.g. `¬cursor.IsTerminal ↔ IsNode cursor.residual`. Without the bridge, downstream code holding a non-terminality fact (the natural hypothesis in reachable-prefix arguments) has to reprove the case split to enter `AppendView.left`. Both are two-liners.

dtumad (commented):
Reviewed the diff. The mathematics is sound and genuinely cast-free as advertised — `join_split`/`split_join` establish the classification as a bijection, the junction boundary case is forced correctly into `right` by the `IsNode` requirement, and the `Decoration.append` universe change is a strict generalization with untouched proof bodies. Two inline items on API placement, plus two smaller ones here:

1. **Missing observational transport lemmas**: the base Cursor layer exposes `plug`/`trace`/`length` as the observational API, but `liftAppend`/`joinRight` get only `residual_*` and constructor simp lemmas. In particular `(joinRight program suffix path cursor).plug q = Path.append program suffix path (cursor.plug q)` is the glue lemma downstream reachable-prefix/composition proofs will ask for first; `trace_liftAppend`/`length_liftAppend` similarly. Fine as a fast-follow if you prefer, but worth landing before this API gets external consumers.

2. **AGENTS.md Module Layering**: the chain at AGENTS.md:100-102 still reads `Free/Cursor → Cursor/Occurrence → Cursor/Fork` and omits `Cursor/Append` (and `Displayed/Cursor` from #58, which this inherits). Since repo-map.md was updated, the mirrored block should be too — probably worth syncing once for the whole stack.

PR #62, opened 2026-07-16
refactor(free): remove phantom telescope facade

quangvdao:
## Summary

- remove the unused `FreeM.TelescopeWith` compatibility facade and its `unusedVariables` / `unusedArguments` linter suppressions
- define canonical-path `FreeM.Telescope` directly as a specialization of `FreeM.StoppingTree`
- preserve the existing `Telescope.done`, `Telescope.extend`, `Telescope.toFreeM`, and computation-law surface
- replace a tautological `FreeP` equality test with a concrete representation canary for `TypeTree.stepPoly`
- update the PFunctor module guide

## Rationale

`TelescopeWith` carried a phantom `round` parameter that did not affect its carrier and required explicit linter suppression. It had no consumers in PolyFun or the checked local downstream repositories. The general abstraction is already `StoppingTree`; the useful canonical-path specialization remains `Telescope`.

This branch was rebuilt directly on top of merged PR #59 so #62 serves as the cleanup PR rather than retaining the former stacked PF-6A feature slice.

## Validation

- `lake exe lint-style`
- `git diff --check`
- targeted build of `PolyFun.PFunctor.Free.Path`, `PolyFun.Interaction.Basic.Telescope`, and `PolyFunTest.Interaction.Basic.FoundationNormalization`
- axiom audit: `Telescope.toFreeM_done` and `Telescope.toFreeM_extend` are axiom-free
- no new `sorry`, `admit`, or `stop`


PolyFun/PFunctor/Free/Path.lean
```diff
@@ -994,39 +994,154 @@ theorem projectPathAlong_append {β : Type t} (l : Lens P Q) :
...
+  | _, .extend s cont => by
+      rw [hExtend]
+      simp only [fold]
+      congr 1
+      funext obs
+      exact eq_fold alg f hDone hExtend (cont obs)
+
+end StoppingTree
+
+set_option linter.unusedVariables false in
```
dtumad: Confirming the Codex P1 with a concrete zero-cost fix: this `set_option linter.unusedVariables false` violates the repo's hard rule against linter suppressions, and it's unnecessary — renaming the phantom binder to `_round` silences **both** linters with zero suppressions (the core linter honors `_`-prefixes, and Batteries' `unusedArguments` linter explicitly allows `_`-prefixed arguments per its own docstring). `round` is never referenced in the abbrev's signature or body, and its only in-file use site passes it positionally, so the rename is free. That also moots the `@[nolint unusedArguments]` below, which currently resolves only transitively — the repo rule asks for a direct `import Batteries.Tactic.Lint` when `@[nolint]` is used.

PolyFun/PFunctor/Free/Path.lean
```diff
@@ -994,39 +994,154 @@ theorem projectPathAlong_append {β : Type t} (l : Lens P Q) :
...
+/-- Compatibility presentation of `StoppingTree` tied to a family of `FreeM`
+rounds. The `round` parameter does not affect the inductive carrier; it only
+records which rounds an observation-based flattening belongs to. New generic
+code should use `StoppingTree` directly. -/
+@[nolint unusedArguments]
+abbrev TelescopeWith {St : Type z} {Out : St → Type v}
```
dtumad: The `TelescopeWith` compatibility facade preserves compatibility for nobody: `git grep TelescopeWith` on head has no occurrences outside this file — its only pre-PR external consumer (`Interaction/Basic/Telescope.lean`) was migrated to `StoppingTree` in this same diff, and even the new test file uses `StoppingTree` directly. The facade costs ~85 lines of wrappers whose docstrings all begin "Compatibility …" (bordering on the repo's ban on change-history docstring wording). Cleanest fix: define `Telescope` directly over `StoppingTree` and delete `TelescopeWith` outright — which also deletes both linter items above. If it must stay for downstream-repo compatibility, keep just the type alias with `_round`.

PolyFunTest/Interaction/Basic/FoundationNormalization.lean
```diff
@@ -0,0 +1,88 @@
...
+Checks that interaction specs, finite chains, and stopping trees expose the
+canonical polynomial structures used by their implementations.
+-/
+
+namespace Interaction
+namespace Spec
+
+/-! ## Specs are the free substitution monoid -/
+
+example : Spec.stepPoly = PFunctor.FreeP Spec.basePFunctor := rfl
```
dtumad: This regression is tautological — `stepPoly` is *defined* as an abbrev of exactly this RHS, so the example cannot fail while the file compiles. The load-bearing regression is the opposite direction of the identification: `example : Spec.stepPoly = ⟨Spec, Transcript⟩ := rfl`, which guards `FreeP` staying reducibly `⟨FreeM P PUnit, Path⟩` — the invariant the whole normalization rests on. (The `succEquiv` re-typecheck below is similarly non-falsifiable; the append/split bridges, `height` and `eq_fold` tests are genuinely falsifiable and good.)

dtumad (commented):
Reviewed the diff. The core normalization is genuine: the identification `Spec.stepPoly ≡ FreeP Spec.basePFunctor` holds through the full reducibility chain (all bridge lemmas are honest `rfl`s), Critical Gotcha #3 is untouched (pre-existing `rfl` tests still pass on head), `StoppingTree.fold`/`eq_fold` is correctly stated and cast-free, and the termination-claim correction fixes a docstring that was demonstrably false (`Telescope.done s₀` inhabits at every state). This is the PR that makes the Interaction layer *inherit* its composition algebra from #54/#55 instead of re-proving it. Three items inline — the first is the merge gate.

PR #64, opened 2026-07-16
Refactor sequential interaction specs as type trees

quangvdao:
## Summary

- rename the bare sequential carrier from Interaction.Spec to Interaction.TypeTree
- rename structural transcripts to TypeTree.Path and migrate the associated Chain, Telescope, sampler, two-party, multiparty, concurrent, and UC APIs
- rename StepOver.spec to StepOver.tree and consolidate concurrent path relations as StepRel
- regenerate the umbrella import and update maintained documentation without compatibility aliases

## Stack

This PR targets main as requested, but is stacked directly on the exact green head of PF-6A (#62), commit 11e02c0101933b7966123017b40cffdfed1399bf. Until #62 and its lower stack merge, GitHub will show those inherited commits in this PR.

The PF-6R commit itself is 6975ff5 and is semantically conservative: the FreeM representations, universes, reducibility, match-pattern constructors, proof terms, and definitional equations are unchanged modulo public names.

## Validation

- ./scripts/validate.sh --lint --test
- lake exe lint-style
- generated umbrella import check
- documentation integrity check
- historical-name and compatibility-alias negative searches
- no new sorry, admit, axioms, unsafe declarations, or linter suppressions

## Independent review

A fresh review-lean-formalization subagent compared every changed Lean source/test file against PF-6A using comment-stripped token streams normalized only by the approved rename map. It found no semantic, statement, universe, abstraction-boundary, or trust-surface change. Two P3 documentation findings were fixed and the final re-review returned no findings.


dtumad (commented):
Verified the rename-only claim on this PR's own commit (6975ff5 vs the #62 head it's stacked on). Method: derived the forward rename map from the diff itself (`Spec→TypeTree`, `Transcript→Path`, `TranscriptRel→StepRel`, `DecoratedSpec→Decorated`, `StepOver.spec→.tree`, ~20 compound forms), applied it to every changed Lean file at the base commit via `sed`, and diffed against the head. Result: residue across all ~57 Lean files is exclusively docstring prose, 100-column rewraps forced by the longer names, and the intentionally-kept `Concurrent.Spec`. Declaration counts are exactly equal (1293 = 1293) across `PolyFun/Interaction` + `PolyFunTest/Interaction`; no statements, universes, attributes, or proof terms changed. In particular:

- The apparent asymmetries are benign: `Chain.lean` is 87/87; `TwoParty/Strategy.lean` (+255/−238) is 100% line rewrapping.
- Gotcha #3 survives: `@[match_pattern, reducible]` intact on `TypeTree.done`/`TypeTree.node`, along with the eliminator attributes and the `abbrev` reducibility of `TypeTree` itself; `AGENTS.md` and `docs/wiki/gotchas.md` restate the invariant correctly.
- `SpecFintype → TypeTreeFintype` is content-complete declaration-by-declaration.
- No lingering old names anywhere (sources, docs, `nolints.json` unchanged), and no compatibility aliases, as claimed.
- The Codex P1 about `set_option linter.unusedVariables false` at `Free/Path.lean:1070` is fully inherited from #62 — this commit doesn't touch that file; it should be resolved there.

One open question for the merge discussion rather than a defect: `Concurrent.Spec` (the structural concurrent inductive) deliberately keeps its name, so `Concurrent.Spec.ofSequential : Interaction.TypeTree → Concurrent.Spec` now mixes both vocabularies in one signature, and `Concurrent/Spec.lean` sits where `Basic/Spec.lean` used to. Defensible as out-of-scope, but worth an explicit decision (follow-up rename to e.g. `Concurrent.Shape`, or a deliberate keep with a naming note). Binder-level polish (`(spec : TypeTree)` survives widely in the Basic layer while Concurrent renamed binders to `tree`) is purely cosmetic and can wait.

Mechanics note: since this targets main while stacked on #62's head, it can merge only after #54/#55/#62 land (at which point its main-diff collapses to exactly this rename commit).

PR #63, opened 2026-07-16
feat(pfunctor): extract category operations from polynomial comonoids

quangvdao:
## Context

A polynomial substitution comonoid encodes a category in outgoing-arrow form:
positions are objects, directions are arrows leaving an object, comultiplication
exposes targets and composable continuations, and the counit selects identities.
The later cofree universal-property and mate results need this producer-side API.

## What this adds

- `Comonoid.identity`, `target`, and `compose`;
- the category laws derived from the comonoid lens equations;
- normalized comultiplication lemmas for dependent target transport;
- `isStateSystem_iff_target_bijective`;
- the retrofunctor laws for `Comonoid.Hom`;
- `Hom.ofCategoryLaws` and composition of comonoid homomorphisms.

## Mathematical boundary

This is the extraction direction of the outgoing-arrow correspondence. It does
not install a `Category` instance on `C.carrier.A`: a comonoid is extra data,
not a property of the bare object type. The reverse category-to-comonoid
construction and a packaged equivalence remain out of scope.

## Coverage

Tests include state comonoids, independent-universe API canaries, a non-thin
one-object Boolean-list monoid, an observable Boolean-negation endomorphism,
and distinct parallel arrows checking composition/path order.

## Stack

- Base: current `main` (through #62).
- This is the first PR in the coherent #63 → #65 → #67 → #68 merge group.
- Next: #65, the cofree polynomial universal property.

## Validation

After restacking onto current `main`:

- targeted source/test build: 739 jobs;
- full build: 8,828 jobs;
- full tests: 1,968 jobs;
- `./scripts/validate.sh --lint --test`;
- `lake exe lint-style`;
- imports/docs integrity and `git diff --check`;
- trust scan with no `sorry`, `admit`, `axiom`, or `unsafe`;
- representative axiom audit: only `propext`, `Classical.choice`, and
  `Quot.sound`.

PR #65, opened 2026-07-16
feat(pfunctor): prove the cofree polynomial universal property

quangvdao:
## Context

#63 extracts the category structure of a polynomial comonoid. This PR proves
the cofree universal property: homomorphisms from a polynomial comonoid `C`
into `CofreeP P` are exactly lenses from `C.carrier` to the generator `P`.

## What this adds

- `CofreeP.cogenerator`;
- generic `unfoldShape`, `unfoldDirection`, and `unfoldLens`;
- public child and append/composition equations;
- `CofreeP.extend` and `restrict` with both triangle laws;
- full uniqueness of the dependent backward maps on finite vertices;
- `CofreeP.homEquiv` and naturality in source and generator.

The result is explicit hom-set equivalence/adjunction data, not a global
`CategoryTheory.Adjunction`. Generic coiteration keeps four independent
universes; packaging through the existing homogeneous `Comonoid.Hom` uses the
documented common maximum.

## Stack

- Base: current `main` through merged #63.
- Previous: #63.
- Next: #67, dynamical-system cofree mates.

## Validation

After restacking onto merged #63:

- targeted source/test build: 749 jobs;
- full build: 8,829 jobs;
- full tests: 1,970 jobs;
- `lake exe lint-style`;
- principal universal-property/naturality axiom audit: only `propext`,
  `Classical.choice`, and `Quot.sound`.

### DIFF
--- a/PolyFun/PFunctor/Cofree/Universal.lean
+++ b/PolyFun/PFunctor/Cofree/Universal.lean
@@ -1,0 +1,821 @@
+/-
+Copyright (c) 2026 PolyFun Contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Quang Dao
+-/
+module
+
+public import PolyFun.PFunctor.Cofree.Polynomial
+public import PolyFun.PFunctor.Comonoid.Category
+
+/-!
+# The universal property of the cofree polynomial comonoid
+
+For a polynomial `P`, the cofree comonoid `CofreeP.comonoid P` is right
+adjoint to the carrier-forgetful functor from polynomial comonoids.  Concretely,
+a lens from a comonoid carrier to `P` extends uniquely to a retrofunctor into
+the cofree comonoid.
+
+This file packages the adjunction at the hom-set level.  It does not install a
+bundled `CategoryTheory.Adjunction`: `PFunctor` currently has overlapping lens
+and chart category instances, whereas the concrete hom-set equivalence is
+unambiguous and is the API needed by downstream dynamical systems. Generic
+coiteration remains universe-polymorphic; only the retrofunctor packaging uses
+the common-maximum universe boundary currently required by `Comonoid.Hom`.
+-/
+
+@[expose] public section
+
+universe uA uB uA₂ uB₂ uCA uCB
+
+namespace PFunctor
+namespace CofreeP
+
+variable {P : PFunctor.{uA, uB}}
+
+/-! ## Cogeneration and coiteration -/
+
+/-- The one-layer projection from the cofree polynomial to its generator.
+It exposes the root position and pulls a generator direction back to the
+corresponding depth-one vertex.  This is `ε_P^{(1)}` in Spivak--Niu. -/
+def cogenerator (P : PFunctor.{uA, uB}) : Lens (CofreeP P) P where
+  toFunA := M.head
+  toFunB tree direction :=
+    .child direction (.root (M.children tree direction))
+
+@[simp]
+theorem cogenerator_toFunA (tree : M P) :
+    (cogenerator P).toFunA tree = M.head tree :=
+  rfl
+
+@[simp]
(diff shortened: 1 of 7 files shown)


PolyFun/PFunctor/Cofree/Universal.lean
```diff
@@ -0,0 +1,821 @@
...
+      Comonoid.compose C object first' second' := by
+  cases hfirst
+  cases hsecond
+  rfl
+
+/-- Pulling back a concatenated cofree vertex composes the arrows represented
+by its prefix and suffix, in outer-then-inner path order. The suffix is
+transported from the selected subtree to the propositionally equal coiterated
+tree at the prefix target. -/
+private theorem unfoldDirection_append (lens : Lens C.carrier P) :
```
dtumad: `unfoldDirection_append` being `private` partially unrealizes the file's own headline generality: `unfoldDirection_root` and `subtree_unfoldShape` publicly expose the identity and target laws of `unfoldLens` at fully independent universes, but the third retrofunctor law is only recoverable at `max uA uB` via `(extend C lens).map_compose`. A consumer with `C : Comonoid.{uCA, uCB}` off the maximum — exactly the clientele the universe-boundary paragraph advertises, e.g. behavior-mate work over `stateComonoid S` — would have to re-prove ~160 lines of cast algebra. Its statement is exactly the public `mapCompose` shape of `Hom.ofCategoryLaws`, so making it public is consistent with the stack's existing API (same consideration for the computation rule `unfoldDirection_child`).

dtumad (commented):
Reviewed the diff against Thm 8.45. The statements are correct and complete: `restrict = cogenerator P ∘ₗ hom.toLens` with `extend` the mate via `M.corec`, uniqueness (`extend_restrict`) genuinely covers the backward maps on all vertices via `unfoldDirection_restrict`, both adjunction triangles are present, the naturality squares are correctly two-sided (the left one contravariant per `Hom.comp`'s diagrammatic order), and the shape-uniqueness proof properly routes through the existing `M.corec_eq_corec` bisimulation infrastructure rather than hand-rolled `M.bisim`. The four-independent-universe claim for generic coiteration checks out and is canary-tested. One API item inline.

Minor naming note: Mathlib's adjunction API spells these `homEquiv_naturality_left/right`; `homEquiv_natural_left/right` is a near-miss if alignment is desired.

The homogeneous `Comonoid.Hom` boundary (`map_counit` is the sole heterogeneity blocker — its two sides live at different `X` universe instances for mixed-universe comonoids, a type mismatch rather than a proof obstacle, while `map_comult` is already heterogeneous-ready) recurs across #65/#67/#69/#70 — flagging the resolution strategy (heterogeneous Hom vs a ULift-comonoid transport functor) as a stack-level decision for @dtumad rather than a condition on this PR.

PolyFun/PFunctor/Cofree/Universal.lean
quangvdao: Addressed in `5549389`, with transport normalization in `11b4282`. `unfoldDirection_child` and `unfoldDirection_append` are public, documented, fully heterogeneous producer laws; off-maximum consumers no longer need to re-prove the child/composition cast algebra.

quangvdao:
Response to @dtumad's review (current head `67d7402`):

- Made `unfoldDirection_child` and `unfoldDirection_append` public, documented, fully heterogeneous producer laws, so off-maximum consumers can reuse the third retrofunctor/composition equation.
- Renamed the hom-equivalence laws to Mathlib-aligned `homEquiv_naturality_left` and `homEquiv_naturality_right`, updating canaries.
- Kept the homogeneous `Comonoid.Hom` limitation explicit without overclaiming; the common explanation is centralized upstream.
- During restacking, normalized the entire cofree universal transport surface to the public `M.Vertex.castEquiv` API, which removed the downstream reliance on raw-cast definitional coincidence.

The public-law inline request has a separate exact reply.

PR #66, opened 2026-07-16
feat(interaction): compose dependent TypeTree chains

quangvdao:
## Summary

- add path-dependent `Chain.then` with explicit round counts
- prove raw and flattened left/right unit laws
- expose the dependent path equivalence through split/join operations
- add `Chain.liftThen` for dependent output families
- add `Chain.strategyCompThen` for strategy composition at the public boundary
- prove operational three-stage associativity
- update the generated imports, interaction wiki, and executable examples

## Semantic contract

A `Chain n` is an intensional finite-round presentation. `Chain.then` runs the prefix for `m` rounds and chooses the `n`-round suffix from the complete prefix path. The theorem `Chain.toTypeTree_then` identifies this operation with generic `PFunctor.FreeM.append`.

Associativity is deliberately stated for the stable operational interpretation, `toTypeTree`, rather than as raw equality between intensional `Chain` presentations. `Chain.toTypeTree_then_assoc` is the public three-stage law.

## Completed client boundary

The former external-client merge gate is superseded by an in-repository, branch-dependent executable example. It exercises:

- a `Bool` prefix branch
- `Fin 2` / `Fin 3` middle branches
- dependent `Fin 12` / `Fin 23` terminal outputs
- path splitting and reconstruction
- `liftThen` computation
- executable `strategyCompThen` path and dependent-output computation

This also closes the earlier deferred `outputFamily` and strategy-composition work against the current `StrategyOver` foundations.

## Validation

Validated at exact head `90beaf92c37317f45538ee7bce80ea5edf42c9e4`:

- `lake exe lint-style`
- `git diff --check`
- axiom audit of the new public declarations (only standard Lean axioms where required)

### DIFF
--- a/PolyFun/Interaction/Basic/Chain/Append.lean
+++ b/PolyFun/Interaction/Basic/Chain/Append.lean
@@ -1,0 +1,347 @@
+/-
+Copyright (c) 2026 PolyFun Contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Quang Dao
+-/
+import PolyFun.Interaction.Basic.Chain
+
+/-!
+# Dependent concatenation of finite interaction chains
+
+This file composes two `TypeTree.Chain` values when the suffix chain may
+depend on the complete path through the prefix. The construction preserves the
+round boundary carried by `Chain`: a chain of `m` rounds followed by a chain of
+`n` rounds produces a chain of `m + n` rounds.
+
+`Chain.then` is the chain-level counterpart of `PFunctor.FreeM.append`.
+`toTypeTree_then` states that flattening commutes with concatenation. The API
+includes raw and flattened unit laws, a path equivalence with inverse split/join
+operations, `liftThen` for dependent output families, and `strategyCompThen`
+for strategies crossing the boundary. `toTypeTree_then_assoc` gives the typed
+three-stage reassociation law and exposes the unavoidable `Nat.add_assoc`
+transport between the two chain indices.
+
+Raw `Chain`-level associativity is intentionally not the public contract:
+it would identify continuation functions through the intensional details of
+round-count and path transports. `Chain` is a finite-approximant presentation,
+while `toTypeTree` is its operational interpretation; consumers therefore use
+the stable operational equality `toTypeTree_then_assoc`.
+-/
+
+universe u
+
+namespace Interaction
+namespace TypeTree
+namespace Chain
+
+/-- Transport a chain across an equality of its round count. -/
+def castRounds {m n : Nat} (h : m = n) (c : Chain m) : Chain n :=
+  h ▸ c
+
+@[simp]
+theorem castRounds_rfl {m : Nat} (c : Chain m) : castRounds rfl c = c :=
+  rfl
+
+@[simp]
+theorem castRounds_symm {m n : Nat} (h : m = n) (c : Chain m) :
+    castRounds h.symm (castRounds h c) = c := by
+  subst n
+  rfl
+
+theorem castRounds_trans {m n p : Nat} (h : m = n) (h' : n = p)
+    (c : Chain m) :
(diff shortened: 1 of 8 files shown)


PolyFun/Interaction/Basic/Chain/Append.lean
```diff
@@ -0,0 +1,215 @@
...
+This file composes two `TypeTree.Chain` values when the suffix chain may
+depend on the complete path through the prefix. The construction preserves the
+round boundary carried by `Chain`: a chain of `m` rounds followed by a chain of
+`n` rounds produces a chain of `m + n` rounds.
+
+`Chain.then` is the chain-level counterpart of `PFunctor.FreeM.append`.
+`toTypeTree_then` states that flattening commutes with concatenation. The path
+API splits and joins a path at the chain boundary, and `toTypeTree_then_assoc`
+gives the typed three-stage reassociation law. The latter exposes the unavoidable
+`Nat.add_assoc` transport between the two chain indices rather than hiding it
```
dtumad: The docstring scopes the associativity claim to flattened trees well, but leaves the raw-Chain question implicit: it never says whether the stronger `reassoc ((c.then k).then l) = c.then …` (Chain-level equality after transport) is false, unknown, or merely deferred as cast-heavy — it's plausibly true, since both sides have identical round trees. One sentence stating the raw law is deliberately not attempted (and why) would answer the reviewer question the PR body anticipates.

PolyFun/Interaction/Basic/Chain/Append.lean
```diff
@@ -0,0 +1,215 @@
...
+-/
+
+universe u
+
+namespace Interaction
+namespace TypeTree
+namespace Chain
+
+/-- Transport a chain across an equality of its round count. -/
+def cast {m n : Nat} (h : m = n) (c : Chain m) : Chain n :=
```
dtumad: Two related notes: (a) the file proves associativity but no unit laws — `toTypeTree_then_zero_right` (composing with the empty chain flattens to `toTypeTree m c`, via `FreeM.append`'s pure-continuation law) and its left counterpart are cheap with the same technique as `toTypeTree_then`, and an associativity-only API is lopsided for the graded-monoid story this file is telling; (b) `Chain.cast` shadows `_root_.cast` inside `namespace Chain`, so a downstream file that `open`s `Chain` may get surprising resolution — Mathlib has `.cast` precedent so this is defensible, but `castRounds` would be unambiguous.

dtumad (commented):
Reviewed the diff. No correctness defects: `Chain.then` is genuinely dependent (suffix selected by the full flattened prefix path), round-count transport is confined to a small `Chain.cast` along `Nat` lemmas with no Fin arithmetic or HEq, `toTypeTree_then` lands exactly on #62's substitution-monoid substrate, and associativity correctly delegates to the pre-existing `FreeM.Path.append_tree_assoc` rather than re-proving it. The casts are computation-transparent (`finalArity … = 12/23 := rfl` reduce through `Equiv.cast`) and the tests pin both prefix branches with distinct arities. Two inline suggestions; one metadata item:

**Draft status**: the PR body says "intentionally a draft" with the ArkLib triple-reduction-client acceptance gate, but the PR is not actually marked draft on GitHub — nothing currently prevents an accidental merge past the self-declared gate. Suggest converting to draft (or a `do-not-merge` label) until the client lands.

dtumad (approved):
Reviewed closely against current main and the current Chain, FreeM, and StrategyOver foundations.

The rebuild now supplies the previously missing semantic boundary: path split/join equivalence, dependent output-family composition, executable strategy composition, unit laws, and operational three-stage associativity. The distinction between intensional Chain presentation and its stable toTypeTree semantics is explicit and appropriate.

I also checked the branch-dependent Bool / Fin 2 / Fin 3 / Fin 12 / Fin 23 example, generated imports/docs, axiom footprint, style, full build, environment lint, and test library. Local validation is green at 90beaf92c37317f45538ee7bce80ea5edf42c9e4.

PR #67, opened 2026-07-16
feat(pfunctor): identify dynamical behavior with cofree mates

quangvdao:
## Context

#65 provides the cofree polynomial universal property. This PR applies it to
dynamical systems and identifies system behavior with the corresponding cofree
mate, including the dependent direction map.

## What this adds

- `unfoldShape_stateComonoid`, connecting generic coiteration to trajectories;
- labeled trajectories and their destructor/head/tail equations;
- `mateObj` with shape and direction decoding;
- `cofreeMate : Comonoid.Hom ...`;
- cogenerator restriction and uniqueness equations;
- the final trajectory/self-label/cofree-mate characterization.

The strengthened state-comonoid API is proved at the owning category layer, and
the examples exercise nontrivial transition/path ordering rather than only thin
or singleton cases.

## Stack

- Base: current `main` through merged #65.
- Previous: #65.
- Next: #68, finite cofree projections.

## Validation

After restacking onto merged #65:

- targeted source/test build: 754 jobs;
- full build: 8,830 jobs;
- full tests: 1,972 jobs;
- `lake exe lint-style`;
- `unfoldShape_stateComonoid` is axiom-free; representative extensional
  results use only `propext`, `Classical.choice`, and `Quot.sound`.


PolyFunTest/PFunctor/Dynamical/CofreeMateExamples.lean
```diff
@@ -0,0 +1,237 @@
...
+    branchingSystem.cofreeMate).toFunB .source false = .middle
+  rw [DynSystem.restrict_cofreeMate]
+  rfl
+
+theorem falseTarget_eq_middle : falseTarget = ThreeState.middle := by
+  unfold falseTarget
+  rw [cofreeMate_falseVertex]
+  rfl
+
+theorem stateComonoid_compose (start first second : ThreeState) :
```
dtumad: This is an instance of general `rfl` facts that don't exist as library lemmas: `Comonoid.identity (stateComonoid S) s = s`, `Comonoid.target (stateComonoid S) s d = d`, `Comonoid.compose (stateComonoid S) s d e = e` (only `stateComonoid_carrier/comult/counit` exist). Every downstream consumer of `cofreeMate` will re-derive these with the same `change`/`rw`/`rfl` gymnastics this test file uses. Suggest adding the three generic `@[simp]` lemmas next to `stateComonoid` — here or in an immediate follow-up.

PolyFun/PFunctor/Dynamical/CofreeMate.lean
```diff
@@ -0,0 +1,285 @@
...
+existing terminal-coalgebra behavior as its object map.  The state and
+interface universes remain independent because this theorem does not package a
+`Comonoid.Hom`. -/
+theorem unfoldShape_stateComonoid (system : DynSystem S P) :
+    CofreeP.unfoldShape (stateComonoid S) system = system.behavior :=
+  rfl
+
+/-! ## Vertex-labelled mate semantics -/
+
+/-- The infinite trajectory of a dynamical system with every reached state
```
dtumad, line 51: Spelling split inside one file: identifiers commit to American `labeledTrajectory`/`labeledBehavior` while the docstrings here and below write "labelled"/"labelling" (and the base-branch test fixture is `labelledTree`). The repo-wide mix pre-exists, but new code should at least be internally consistent — since the identifiers are already "labeled", aligning the docstrings is the cheap direction.

dtumad (commented):
Reviewed the diff. The central claims all verify: `stateComonoid S` is the genuine codiscrete monomial comonoid `Sy^S`, `unfoldShape_stateComonoid` is a `rfl` theorem so the existing `DynSystem.behavior` is *identified* with coiteration rather than shadowed (`Trajectory.lean` is untouched), trajectory recovery is definitional, all seven `cast`/`HEq` occurrences sit inside three `private` theorems with every public statement cast-free (as the PR body claims), and the universe collapse to `max uA uB` is confined to the `Mate` section where the homogeneous `Comonoid.Hom` forces it. The canaries pin the dependent backward map at root, both one-edge branches, and a two-edge composite. Three small items inline.

One robustness note (non-blocking): the `rfl` bridges typecheck because `Comonoid.target`'s cast reduces via proof irrelevance against `fixState`'s literal shape — a comment noting the definitional dependence (and perhaps a pointer back from `Trajectory.lean`) would guard future refactors from silently breaking the identification.

PolyFun/PFunctor/Dynamical/CofreeMate.lean
```diff
@@ -0,0 +1,285 @@
...
+
+private theorem mateObj_child
+    (system : DynSystem S P) (state : S) (label : S → α)
+    (direction : P.B (M.head (mateObj system state label).1)) :
+    CofreeP.childObj (mateObj system state label) direction =
+      mateObj system (system.update state
+        (cast (congrArg P.B
+          (CofreeP.head_unfoldShape (stateComonoid S) system state))
+          direction)) label := by
+  let childEq := CofreeP.children_unfoldShape
```
dtumad: `let childEq := …` (and `let hPosition := by …` below) bind Prop-valued proof terms; Mathlib style uses `have` for proofs — `let` needlessly retains the body in the term.

PolyFunTest/PFunctor/Dynamical/CofreeMateExamples.lean
quangvdao: Addressed in `c14b6c`. I added generic `[simp]` lemmas for identity, target, and composition of `stateComonoid` at the first layer where those category operations exist, then removed the local test-only rederivation.

PolyFun/PFunctor/Dynamical/CofreeMate.lean
quangvdao: Addressed in `c14b6c` and completed at the owning #67 layer in `c3744e6`. All new `CofreeMate` prose now uses American `labeled`/`labeling`, matching the public identifiers.

PolyFun/PFunctor/Dynamical/CofreeMate.lean
quangvdao: Addressed in `c14b6c`. The Prop-valued `childEq`/`hPosition` bindings were changed from `let` to `have`, so proof bodies are not retained unnecessarily.

quangvdao:
Response to @dtumad's review (current head `c3744e6`):

- Added generic `[simp]` laws for identity, target, and composition of `stateComonoid` at the category-operation layer and removed the test-local rederivation.
- Changed Prop-valued proof bindings from `let` to `have`.
- Added documentation calling out the definitional/proof-irrelevance dependency behind the `rfl` state-comonoid bridges.
- Standardized all new `CofreeMate` prose on American `labeled`/`labeling`; the last later occurrence was fixed at this owning PR before this response.

Each inline thread has its corresponding implementation reply.

PR #68, opened 2026-07-16
feat(pfunctor): formalize finite cofree projections and Proposition 8.49

quangvdao:
## Context

#67 identifies a dynamical system with its infinite cofree mate. This PR exposes
the exact finite-depth observations of that object as lenses into iterated free
polynomials, formalizing the finite projection result corresponding to
Proposition 8.49.

## What this adds

- `CofreeP.projectionN` for every finite depth;
- zero/successor component equations and depth behavior;
- one- and two-step diagrammatic characterizations;
- the general `projectionN_eq_comultN_compNthMap` equation;
- naturality with respect to comonoid homomorphisms and cofree extension;
- `DynSystem.cofreeMate_comp_projectionN` and its two-step specialization.

Tests cover concrete finite truncations, direction transport, composition order,
and the dynamical-system specialization. Diagrammatic projection notation is
registered explicitly in the style allowlist.

## Stack

- Base: current `main` through merged #67.
- Previous: #67.
- This closes the coherent #63 → #65 → #67 → #68 merge group.

## Validation

After restacking onto merged #67:

- targeted source/test build: 759 jobs;
- full build: 8,832 jobs;
- full tests: 1,976 jobs;
- `lake exe lint-style`;
- representative projection/mate axiom audit: only `propext`,
  `Classical.choice`, and `Quot.sound`.


PolyFun/PFunctor/Cofree/FiniteProjection.lean
```diff
@@ -0,0 +1,209 @@
...
+                (comonoid P).comultN n) ∘ₗ comult) := by
+        simpa only [comonoid_carrier, comonoid_comult] using hassoc
+      exact houter'.trans hassoc'
+
+/-- A retrofunctor into the cofree comonoid commutes with every finite
+projection.  This is the structural form of the finite-run equation, with the
+one-layer restriction of the retrofunctor on the right. -/
+theorem hom_comp_projectionN
+    {P : PFunctor.{u, u}} {C : Comonoid.{u, u}}
+    (hom : Comonoid.Hom C (comonoid P)) (n : ℕ) :
```
dtumad, line 174: Naming-direction inconsistency: all six new `_comp_` theorems here and in `Dynamical/CofreeMate/FiniteProjection.lean` are named in `⨟` (diagrammatic) order while their statements are written in `∘ₗ` order — e.g. `hom_comp_projectionN` states `projectionN P n ∘ₗ hom.toLens = …`. That inverts the convention of both main (`sumMap_comp_inl : (l₁ ⊎ₗ l₂) ∘ₗ Lens.inl = …`) and the sibling file this PR imports (`cogenerator_comp_unfoldLens : cogenerator P ∘ₗ unfoldLens C lens = lens`). Either rename (`projectionN_comp_hom`, …) or restate in `⨟` so name and notation agree — given the pending ⨟-unification refactor, restating in `⨟` (as `nStep` itself is defined) may be the forward-compatible choice, but it should match #67 either way.

dtumad (commented):
Reviewed the diff. `projectionN` is the correct structural realization of the trimming projection `ε_p^{(n)}` under the repo's right-nested `compNth` convention, with both lens components pinned by `rfl` simp lemmas and exact depth-`n` backward semantics; the three-tier Prop 8.49 (arbitrary retrofunctor → cofree extension → `cofreeMate`/`nStep`) is the full-strength statement, and the Eq 8.32 deferral is sound — 8.32 is the duplicator's additive-splitting spec from §8.1.2, independent of 8.49, and its two sides aren't even type-equal under right-nested `compNth` without the reassociation equivalence (the same all-bracketing canonicity gap the roadmap already tracks for Prop 7.20(d)). Proofs are cast/HEq-free and genuinely reuse #67's machinery and main's `map_comultN`/`compMap_comp`. Tests pin ordered two-edge behavior (a transposed backward map fails). One naming item inline; one robustness note:

`cofreeMate_comp_projectionN`'s proof (`exact CofreeP.extend_comp_projectionN …`) relies on silent definitional unfolding of both `cofreeMate` and `nStep`; routing through the stated equations with a `rw`/`show` would survive a refactor of either def to a structure-eta-blocking form. Optional.

PolyFun/PFunctor/Cofree/FiniteProjection.lean
quangvdao: Addressed in `b78d36f` (current PR head `155493b`). All six `_comp_` equations are stated in diagrammatic `⨟` order, aligning theorem names, notation, and the `nStep` convention. The established notation locations are registered in the repository style baseline.

quangvdao:
Response to @dtumad's review (current head `155493b`):

- Restated all six `_comp_` equations in diagrammatic `⨟` order, so theorem names, notation, and the `nStep` convention agree.
- Reworked `cofreeMate_comp_projectionN` to pass through the stated `cofreeMate_eq_extend` and `nStep_eq` equations rather than relying on silent definitional unfolding.
- Registered these new source uses of the repository's established diagrammatic notation in the existing Unicode style baseline.

The naming-direction inline thread has the exact commit-level response.

PR #69, opened 2026-07-16
feat(pfunctor): construct tensor comonoids and convolution monoids

quangvdao:
## Context and motivation

Libkind–Spivak constructs the universal interaction map by currying a generator into an internal hom `[C, M]`. For that internal hom to carry a substitution-monoid structure, PolyFun needs the concrete duoidal interaction between tensor product and polynomial substitution, tensor products of comonoids, and the associated convolution construction.

This PR supplies that algebraic layer. It is the last prerequisite before the cofree laxator and `Xi`.

### Duoidal coherence

- the two interchange associativity diagrams;
- all four mixed-unit diagrams;
- tensor associator and unitor naturality needed by the construction;
- the shared unit as both a tensor monoid and substitution comonoid.

Generic tensor/unit coherence lives in `Lens/Basic`; `Lens/Duoidal` is kept focused on the interaction between `⊗` and `◃`.

### Tensor comonoids

- heterogeneous tensor products of polynomial comonoids;
- tensor products of retrofunctors;
- pointwise tensor unitors and associator;
- counit/comultiplication preservation and coherence laws.

### Convolution substitution monoids

For `C : Comonoid` and `M : SubstMonoid`:

- the internal hom `C ⊸ M`;
- raw unit and multiplication maps with exact uncurrying equations;
- `SubstMonoid.convolution C M`;
- left/right unit and associativity laws derived from the duoidal diagrams.

The exact universe pair is:

```lean
SubstMonoid.{max cA cB mA mB, max cA mB}
```

This is the universe of the polynomial internal hom, not an arbitrary collapse.

## API ownership and proof design

Review moved reusable facts to their true owners:

- `Lens.Equiv.cancel_toLens` is in `Lens/Basic`;
- uncurry injectivity uses `Lens.curryEquiv.symm.injective`;
- `unitComparison`, `xTensor_natural`, and `tensorX_natural` are tensor facts in `Lens/Basic`.

`convolutionUnitRaw` must repeat two private helper bodies because a public exposed definition cannot reference those private names under the module visibility boundary. The source records this and the definitional-equality dependency used by the factorization proofs.

## Falsifiable coverage

The tests exercise:

- all six coherence diagrams;
- independent universe combinations;
- tensor-comonoid backward maps;
- convolution unit and multiplication;
- order-sensitive state threading;
- branch-dependent continuations, including both branches.

The branch-dependent example was added specifically so a wrong continuation choice could not pass under constant data.

## Stack and source context

- Stack position: **9 of 12**.
[...]

PR #70, opened 2026-07-16
feat(pfunctor): prove the cofree polynomial lax-monoidal structure

quangvdao:
## Context and motivation

The tensor/convolution algebra in #69 is useful only after the cofree polynomial construction is shown to interact coherently with tensor product. Libkind–Spivak uses exactly this lax-monoidal structure to define the convolution monoid into which the pattern-action generator is extended.

This PR constructs the concrete unit and tensor comparison maps for `CofreeP` and proves their naturality and coherence.

### Universal-property extensionality

- `CofreeP.hom_ext_cogenerator`: two homomorphisms into a cofree polynomial comonoid are equal when they agree after the cogenerator.

This lets the large coherence diagrams be proved at the generator layer rather than by redoing dependent M-tree/vertex calculations.

### Unit comparison

- the retrofunctor/lens from the tensor unit into the cofree polynomial on the unit;
- its cogenerator equation;
- left and right unit coherence.

### Tensor comparison

- the map combining a `CofreeP P` tree and a `CofreeP Q` tree into a `CofreeP (P ⊗ Q)` tree;
- its dependent backward map, which splits a finite combined vertex into synchronized source vertices;
- generator equation;
- naturality in both generator lenses;
- associativity coherence.

Together these equations are the concrete lax-monoidal structure needed by the later module action.

## Source and mathematical boundary

This formalizes Libkind–Spivak Proposition C.1 and the concrete Spivak–Niu cofree laxator (current Proposition 8.79; earlier numbering 8.81).

The laws are explicit retrofunctor/lens equations. This PR does not package a general monoidal functor object or a global natural-isomorphism hierarchy.

Naturality is stated within each fixed-universe polynomial category, matching the current homogeneous `Comonoid.Hom` boundary. Fully cross-universe naturality would require a heterogeneous retrofunctor/coiteration-fusion API. That restriction is documented; the underlying generator operations remain separately universe-tested.

## Falsifiable coverage

Tests include:

- unit comparison at observable roots;
- tensor comparison on asymmetric trees;
- separated-universe API canaries;
- naturality equations;
- left/right unit and associativity;
- a decisive depth-two backward-path example that records both synchronized source traversals.

A shape-only construction, swapped tensor component, or incorrect vertex pullback fails these canaries.

## Stack and review history

- Stack position: **10 of 12**.
- Base: #69.
- Next: #71, the pattern-runs-on-matter action.
- Current head incorporates the fully reviewed upstream train; no PR in the train has been merged.

Ready for review; open and unmerged.

## Validation

Targeted source/test builds, the full validation wrapper, style/environment lint, imports/docs checks, diff/trust scans, and principal axiom checks pass.

## Deliberately deferred

The actual action `Xi`, its free universal characterization, and its unit/associativity module laws are #71.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -133,6 +133,8 @@ PFunctor/{Cofree, Comonoid, M/Vertex}
 PFunctor/{Cofree/Polynomial, Comonoid/Category}
   → PFunctor/Cofree/Universal
   → PFunctor/Cofree/FiniteProjection
+PFunctor/{Cofree/Universal, Comonoid/Tensor}
+  → PFunctor/Cofree/LaxMonoidal
 
 PFunctor/{Lens, Cofree, M} + Control/Coalgebra
   → PFunctor/Dynamical/{Basic, Safety, Combinators, Run, Speedup, Trajectory}
--- a/PolyFun.lean
+++ b/PolyFun.lean
@@ -118,6 +118,7 @@ import PolyFun.PFunctor.Category
 import PolyFun.PFunctor.Chart.Basic
 import PolyFun.PFunctor.Cofree
 import PolyFun.PFunctor.Cofree.FiniteProjection
+import PolyFun.PFunctor.Cofree.LaxMonoidal
 import PolyFun.PFunctor.Cofree.Polynomial
 import PolyFun.PFunctor.Cofree.Universal
 import PolyFun.PFunctor.Comonoid
--- a/PolyFun/PFunctor/Cofree/LaxMonoidal.lean
+++ b/PolyFun/PFunctor/Cofree/LaxMonoidal.lean
@@ -1,0 +1,351 @@
+/-
+Copyright (c) 2026 PolyFun Contributors. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Quang Dao
+-/
+module
+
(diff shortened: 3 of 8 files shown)


dtumad (approved):
Reviewed the cofree unit and tensor construction via the universal property. Naturality, unit, and associativity reduce correctly by cogenerator extensionality, and the concrete synchronized-tree tests make direction ordering observable. The unique PR slice is semantically clean; it should be range-diffed again after the parent stack is reconstructed.

dtumad (approved):
Re-reviewed closely on the clean one-commit range. The comparison directions match Spivak–Niu Prop. 8.79 and Libkind–Spivak Prop. C.1: y → T y and T p ⊗ T q → T(p ⊗ q). Naturality has the correct bifunctorial square; both unitor diagrams and the associator diagram have the correct source/target orientation. Defining the maps by cofree extension and proving equality by cogenerator extensionality is semantically appropriate and gives usable Hom- and Lens-level APIs. The synchronized-tree tests observe factor order and two layers of pulled directions, so they rule out copied, swapped, reversed, and root-only implementations. Universe restrictions are explicit and consistent with the current homogeneous Comonoid.Hom boundary. Full local validation passed: 8,850-job build, import/docs checks, environment lint, and the 2,002-job PolyFunTest library. Patch is exact under range-diff.

PR #76, opened 2026-07-17
feat(pfunctor): complete resumption coinduction and lens transport

quangvdao:
## Summary

- add computational-view `HeadMatch`, public `Resumption.bisim`, and `corec_unique`
- add universe-polymorphic interface transport with identity, composition, map, and bind laws
- prove `FreeM.toResumption` naturality under polynomial lenses
- add branch-sensitive and separated-universe producer canaries
- register the coinduction and lens-transport API in the module documentation

## Design boundary

This is the state-free Resumption foundation for #32. It remains independent of ITree, bounded-depth semantics, and stateful `DynComputation` operations.

## Validation and review

- exact repaired head: `882a1d6c28771f4385cefc324881a9f2a88b3dd6`
- focused owning-module and producer-example compilation
- full `validate.sh --lint --test`, text-style lint, diff hygiene, and principal-theorem axiom audit in the review cycle
- independent strict formalization audit; the documentation-registration finding was repaired

Stack root; based on `main`.

PR #77, opened 2026-07-17
feat(itree): characterize tau-free resumptions

quangvdao:
## Summary

- embed computational resumptions into ITrees with bind, map, and lens naturality
- define tau-freeness as the greatest pure/query invariant, admitting infinite visible behavior
- construct the proof-carrying inverse and package the exact equivalence
- keep the inverse constructive and executable; the impossible tau branch uses ordinary `False.elim`
- add branch-sensitive, universe-polymorphic, nullary-query, infinite-loop, and tau-rejection canaries

## Validation and review

- exact repaired head: `fddc6e665df99ded7a8a5112def140ff8bf24e64`
- focused owning-module and producer-example compilation
- full build/lint/test/style/diff validation in the review cycle
- independent strict formalization audit; constructivity wording and DAG/file-index registration findings were repaired

Part of #32. Stacked on #76.


dtumad (approved):
Reviewed the TauFreeF/TauFree characterization and the Resumption equivalence. It rejects exactly trees with a reachable Tau while admitting infinite visible behavior, and the conversions preserve the coinductive structure. No semantic or usability blocker in this slice; please range-diff after #76 is rebased.

dtumad (approved):
Re-reviewed as the isolated fddc6e6 slice rebuilt on current main. Resumption.toITree preserves pure/query heads, bind, map, and lens transport without inserting Tau. ITree.TauFree is the greatest postfixed pure/query invariant: it admits infinite visible-query behavior, recursively requires every query child, and rejects any reachable step. The proof-carrying inverse observes shape constructively, eliminates the impossible Tau branch via False.elim, and the two coinductive round-trip theorems establish the exact subtype equivalence rather than merely an embedding. Branch-sensitive, nullary-query, infinite-loop, Tau-rejection, and universe canaries cover the important edges. The replay applied without conflict and the range diff preserves the original source/tests. Full validation passed: 8,877 build jobs, 2,042 tests, 216 imports, docs integrity, and environment lint. Approving the current head.

PR #124, opened 2026-08-08
feat(interaction): plug factorization for the process model up to activation equivalence

### LINKED ISSUE #123: feat(interaction): plug/par and plug/wire factorization up to activation equivalence
## Problem

`Observation.RespectsFactorization` (added in #121) is not instantiable for `openTheory`. Of the three `OpenTheory.HasPlugWireFactor` fields, only `plug_eq_wire` has an activation-equivalence counterpart (`openTheory_plug_eq_wire_activation_equiv`); `plug_par_left` and `plug_wire_left` do not, and neither do any of the three `IsTraced` laws that would be the categorical route to them. So the process model currently gets the `plug` half of the composition suite via `RespectsPlugComm` but not the `par` / `wire` half.

## Proposed work

Prove `openTheory_plug_par_left_activation_equiv` and `openTheory_plug_wire_left_activation_equiv` directly with `OpenProcessActivationEquiv.of_step_match`, modelled on `openTheory_par_assoc_activation_equiv` — both are re-encodings of a two-bit scheduler path into a one-bit one. Then add the `RespectsFactorization` instance for `openTheory` in `OpenProcessEmulates.lean`, alongside the existing `RespectsPlugComm` one, and extend the right-hand variants from the left ones using `openTheory_par_comm_activation_equiv` and `openTheory_plug_comm_activation_equiv`.

Scouted for the `par` case, which typechecks as far as it goes:

* state relation `fun ⟨⟨s₁, s₂⟩, k⟩ ⟨s₁', k', s₂'⟩ => s₁ = s₁' ∧ s₂ = s₂' ∧ k = k'`, with both totality witnesses by `rfl`
* path correspondence `⟨true, ⟨true, tr⟩⟩ ↔ ⟨true, tr⟩`, `⟨true, ⟨false, tr⟩⟩ ↔ ⟨false, ⟨false, tr⟩⟩`, `⟨false, tr⟩ ↔ ⟨false, ⟨true, tr⟩⟩`
* the `true`/`true` sub-case of `forwardSilent` goes through with the `par_assoc` idiom: `simp only [IsSilentStep, ProcessOver.interleave, Decoration.map]` then an `isSilentDecoration_iff_map` chain discharged by `simp [OpenNodeContext.close, BoundaryAction.closed]` and `simp [OpenNodeContext.inlTensor, BoundaryAction.embedInlTensor]`

The remaining sub-cases are where the work is. Unlike `par_assoc`, the right-hand side carries two `OpenProcess.mapBoundary` wrappers (the `tensorEmptyRight` reshapes that `HasPlugWireFactor.plug_par_left` needs to state at all), and these do not appear as `Decoration.map` — they elaborate to `StepOver.mapContext (OpenNodeContext.map ...)`, so the `isSilentDecoration_iff_map` chain does not line up with the map nesting on the left. Either a `StepOver.mapContext` analogue of `isSilentDecoration_iff_map`, or a normalization lemma turning `mapBoundary` into `Decoration.map` at the step level, would make these mechanical.

## Location

* `PolyFun/Interaction/UC/OpenProcessModel.lean` (the two theorems, next to `openTheory_plug_comm_activation_equiv`)
* `PolyFun/Interaction/UC/OpenProcessEmulates.lean` (the instance)

## Acceptance criteria

* `openTheory_plug_par_left_activation_equiv` and `openTheory_plug_wire_left_activation_equiv` are proved, sorry-free.
* `Observation.RespectsFactorization` has an instance for `openTheory` given `Observation.IsSchedulingInsensitive`, and the "What is still missing" section of `OpenProcessEmulates.lean`'s module docstring is removed.
* `Emulates.par_compose` and `Emulates.wire_compose` are shown applying to `openTheory` in `PolyFunTest/Interaction/UC/EmulatesFactorizationExamples.lean`.
* `OpenProcessModel.lean` stays under the 1500-line cap, or the activation-equivalence laws move to their own module.

dtumad:
Stacked on #121 — review that first; this PR's base is its branch. Closes #123.

## Summary

- add `Interaction/UC/OpenProcessFactorization.lean` with the four factorization laws for `openTheory` up to `OpenProcessActivationEquiv`: `openTheory_plug_{par,wire}_{left,right}_activation_equiv`
- move `openTheory_plug_par_left_activation_equiv` and its siblings into that module rather than growing `OpenProcessModel.lean` past the 1500-line cap (it stays at its current 971 lines)
- add `respectsFactorization_of_isSchedulingInsensitive`, completing the instance story begun in #121
- extend `EmulatesFactorizationExamples.lean` with `par_compose` and `wire_compose` on the process model

## Why

#121 made the composition theorems depend on the observation rather than on strict compact-closed structure, but `openTheory` could still only satisfy the weaker `RespectsPlugComm`. The `par` and `wire` half needed activation-equivalence versions of `plug_par_left` and `plug_wire_left`, which did not exist — and neither did any of the three `IsTraced` laws that would be the categorical route to them.

So they are proved directly, as `OpenProcessActivationEquiv.of_step_match` bisimulations in the established style of `openTheory_par_assoc_activation_equiv`. In each case the state relation is a regrouping of the nested product of component states, and the four step obligations re-encode a two-bit scheduler path as a one-bit one. What makes this work is that the scheduler nodes being moved are `.internal`, so `isActivated` is false and `activationLTS` labels them silent — the delay bisimulation absorbs exactly the nodes the regrouping introduces or removes.

The `_right` variants are proved directly rather than derived from the `_left` ones. The strict proofs of `close_par_right` / `close_wire_right` rewrite under `plug` using `par_comm` / `wire_comm`, which needs activation equivalence to be a congruence for `plug`; no such lemma exists, and `wire_comm` has no activation-equivalence version either. A `plug`-congruence lemma would be independently useful and would shorten this file, but it is a separate piece of work.

The payoff is that the concrete, probability-carrying model now meets the same interface as the free syntax models, so `Emulates.par_compose` and `Emulates.wire_compose` apply to it — which is the general validation that #121's design actually works, not just on the models that already had strict coherence.

**Deliberately excluded.** No `IsTraced` or `IsCompactClosed` instance for `openTheory`; those laws remain unproved for this model and are not needed here. No probability — that stays downstream in VCVio, per the scope boundary.

## Validation

- full library build with warnings as errors
- environment lint, full `PolyFunTest` build, import and documentation integrity checks
- `Obs.RespectsFactorization` resolves by synthesis for `openTheory`, and `par_compose` / `wire_compose` elaborate against it (new examples in the test file)
- all three UC modules stay under the 1500-line cap: `OpenProcessFactorization.lean` 916, `OpenProcessModel.lean` 971, `OpenProcessEmulates.lean` 127
- `docs/wiki/interaction.md` UC table and `docs/wiki/repo-map.md` layering updated for the new module


dtumad:
Note on CI: `ci.yml`, `check-imports.yml`, and `docs-integrity.yml` all trigger only on `pull_request: branches: [main]`, so while this PR is based on #121's branch only Lint Style and the OpenRouter key check run. The full suite ran locally — `./scripts/validate.sh --lint --test` passes, covering the same build/lint/test/import/docs jobs. Retargeting to `main` after #121 merges will run them in CI.

PR #131, opened 2026-08-12
refactor: expose operation-first ITree carrier

dtumad:
## Summary

- define ITree.Poly F α reducibly as F + C α + X
- represent ITree as a one-field wrapper around the raw M-type, with definitionally inverse ofM/toM maps and equivM
- retain the ergonomic Shape API through ViewPoly and migrate corecursion, coinduction, simulations, and event interpreters to it
- align Resumption and DynComputation with the operation-first p + C β convention
- add representation canaries and update the ITree and polynomial-functor documentation

This follows the representation pattern explored in leanprover/cslib#731 while retaining the current cslib dependency pin.

## Validation

- ./scripts/validate.sh --lint --test


quangvdao (approved):
Reviewed exact head `db85108c002af2b4cda9a4ecbe7b6401e85f80e2` after rebasing onto current `main` (`f931c73b4305d9a9f589b6444376640ae706db95`).

The review found one merge-blocking integration defect: `PolyFun.Realizability.Machine.head` still treated the raw `ITree` carrier as return-first. That broke the full library build once the carrier became operation-first. The branch now defines the public return-first readout through the computational `view`, proves its relationship to the raw operation-first position via `Sum.swap`, and updates the `mapResult`, `wrap`, and closure proofs accordingly.

Coverage added and checked:
- concrete `pack`/`unpack`, `viewEquiv`, `shape`/`ofShape`, immediate `corec`, and nontrivial `corec_eq_corec` canaries;
- operation-first Resumption return/query injection canaries;
- existing Realizability head-ordering and representation examples;
- full `./scripts/validate.sh --lint --test --axioms`: build, imports, docs, linters, and tests pass; axiom sweep covers 8,981 declarations across 243 modules with zero `sorry` or non-standard-axiom taint;
- full VCVio downstream build against this PolyFun head passes (2,979 jobs). VCVio required an explicit `BigOperators` import, but the identical failure reproduced against current PolyFun `main`, so it is pre-existing downstream import hygiene rather than a regression from this PR.

No unresolved correctness, universe, abstraction-boundary, executable-behavior, or downstream compatibility findings remain. Approved.

PR #138, opened 2026-08-21
feat(control): exact monadic support over core's MonadAttach

dtumad:
## Summary

Builds PolyFun's support layer on Lean core's `MonadAttach` (shipped since v4.28) rather than on a bespoke class.

- fix an import leak in `Control/Monad/Hom.lean`: drop the unused public `Mathlib.Probability.ProbabilityMassFunction.Monad` (a scope violation per gotcha #3) and `Mathlib.CategoryTheory.Monad.Types`; test files and `Realizability/Instances.lean` that had been receiving `norm_num` and `Fintype` instances through that transitive closure now import what they use
- add `ExactMonadAttach`: the two *introduction* rules for `MonadAttach.CanReturn` that core omits
- add `MonadAttach.support : m α → Set α`, definitionally `CanReturn`, so `a ∈ support x ↔ CanReturn x a` is `Iff.rfl`
- add the judgments `AllOutputs` / `SomeOutput` / `NoOutput` with scoped notation `x ⊨ₐ p` / `x ⊨ₛ p` / `x ⊭ p`, and the demonic `MAlgOrdered m Prop` instance identifying "always" with the trivial-precondition triple
- add `MonadAttach`/`ExactMonadAttach` for `FreeM P` with a computable, axiom-free `attach`, plus `Except` and `SetM` instances (core has neither)

## Why not a bespoke class

Core proves only the *elimination* half of the support theory (`canReturn_bind_imp'`, `eq_of_canReturn_pure`, `canReturn_map_imp'`), which bounds the support from above. That does not pin it down — `fun _ => PUnit` with `CanReturn := False` satisfies `LawfulMonadAttach` vacuously, since every core law is an implication *out of* `CanReturn`. `ExactMonadAttach` supplies the missing introduction rules, so each of core's implications becomes the equivalence support reasoning rewrites with; extending `LawfulMonadAttach` rather than the weak class simultaneously excludes `MonadAttach.trivial` (`CanReturn := True`, i.e. `support = univ`).

The payoff beyond deleting ~120 lines of boilerplate: core supplies the `Id`, `Option`, `OptionT`, and `ExceptT` instances; `FreeM`'s structural equations hold by `rfl`; its value universe is no longer pinned to the direction universe; `MonadAttach.pbind` becomes available for well-founded recursion over free programs; and (in the stacked PR) `WPSound` converts `mvcgen` proofs into support facts.

## Notes

- Core's `MonadAttach (ExceptT ε m)` is declared at `max`-joined universes and cannot be synthesized in a polymorphic context; a single-universe alias works around it and should be deleted once upstream is repaired.
- `StateT`/`ReaderT` keep core's canonical support (the union over initial states), so the elimination theory applies, but they are correctly *not* `ExactMonadAttach` — the tests now *prove* both introduction rules fail there, where the previous design only asserted it in prose.

## Validation

`lake build`, `lake lint`, `lake test`, `./scripts/validate.sh --axioms` (zero-debt baseline preserved). `PolyFunTest/Control/MonadAttach.lean` pins the `Iff.rfl` transfer contract, the `StateT` counterexamples, and `pbind`.

Part 1 of 2; the free-monad wp layer and the `Std.Do`/`mvcgen` bridge build on this in the follow-up PR.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,7 +95,9 @@ and depend on this library.
   and concrete complexity classes live downstream.
 - `PolyFun/Control/`: monad and comonad infrastructure transitively
   required by the above (coalgebra, comonad, free / freecont monad
-  algebra, monad iter / hom, lawful re-exports).
+  algebra, monad iter / hom, lawful re-exports), plus exact monadic
+  support over core's `MonadAttach` and the always/some/never judgments
+  (`Monad/Support`).
 - `PolyFun/Control/LTS/Trace.lean`: generic finite visible traces over the
   silent/visible `Control.LTS` layer and preservation by weak simulation.
 - `PolyFun/Logic/`: small logic helpers (`HEq`).
--- a/PolyFun.lean
+++ b/PolyFun.lean
@@ -12,6 +12,7 @@ public import PolyFun.Control.Monad.FreeCont
 public import PolyFun.Control.Monad.Hom
 public import PolyFun.Control.Monad.Indexed
 public import PolyFun.Control.Monad.Iter
+public import PolyFun.Control.Monad.Support
 public import PolyFun.Control.Trace
 public import PolyFun.IPFunctor.Basic
 public import PolyFun.IPFunctor.Chart.Basic
@@ -194,6 +195,7 @@ public import PolyFun.PFunctor.Free.Path.Execution
 public import PolyFun.PFunctor.Free.Polynomial
 public import PolyFun.PFunctor.Free.Replicate
 public import PolyFun.PFunctor.Free.Resumption
+public import PolyFun.PFunctor.Free.Support
 public import PolyFun.PFunctor.Free.Universal
 public import PolyFun.PFunctor.Handler
 public import PolyFun.PFunctor.Handler.Free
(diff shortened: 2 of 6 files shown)


quangvdao (approved):
Reviewed the complete main...e837a10 change surface with the Lean formalization audit checklist.

The blocking issue was the generic global `MAlgOrdered m Prop` support instance: it overlapped PolyFun's existing `OptionT` / `ExceptT` transformer algebras while assigning failure a different meaning (vacuous partial correctness versus `⊥`). The branch now exposes this semantics as the explicit named `mAlgOrderedPropDemonic` definition and pins the support theorems with a local instance, so downstream inference cannot silently select an incoherent algebra.

I also added producer-level canaries for both FreeM responses, an answer-dependent two-query continuation, `attachSupp` erasure, empty Option/Except support, multi-result SetM support, OptionT/ExceptT success and failure bind paths, and the intentional distinction between default transformer failure and explicit support semantics. Module ownership/docs were updated accordingly.

Exact-head local validation passed: `./scripts/validate.sh --lint --test --axioms`; 8,806 declarations across 239 modules, with zero sorry or non-standard-axiom taint. No remaining correctness, statement, abstraction-boundary, universe, executable-behavior, or proof-debt findings. Approved pending the hosted checks already in progress.

PR #146, opened 2026-08-22
refactor(control): tighten FreeCont transformer API

dtumad:
## Summary

- retain the genuinely distinct Church-encoded free monad transformer while replacing its reducible function alias with a one-field structure
- add the missing effect construction and elimination surface: liftBind, liftF, run equations, extensionality, and ordinary do-notation usability alongside lawful base-monad lifting
- remove the duplicate local FreeM, inert polynomial aliases, and derived bridge conveniences
- keep the essential Cslib.FreeM conversions and one-way round trip, with explicit documentation that no equivalence follows without parametricity
- add a regression example combining an Option base action with a custom effect and executing it through the CPS handler

## Why this shape

Cslib.FreeM provides inspectable inductive syntax but no corresponding free monad transformer. The retained API is therefore useful for direct CPS composition and interpretation. The removed polynomial specialization had no operations or consumers, while the raw type alias prevented Lean from recognizing annotated do blocks as monadic programs.

## Validation

- ./scripts/validate.sh --lint --test --axioms
- 8,787 declarations across 238 PolyFun modules
- zero sorryAx or non-standard axiom taint

### DIFF
--- a/PolyFun/Control/Monad/FreeCont.lean
+++ b/PolyFun/Control/Monad/FreeCont.lean
@@ -6,98 +6,102 @@ Authors: Quang Dao
 module
 
 public import PolyFun.Control.Monad.Free
-public import PolyFun.PFunctor.Free.Basic
 public import Mathlib.Control.Monad.Cont
 
 /-!
-# Continuation-passing (Church-encoded) free monads
+# Continuation-passing free monad transformer
 
-Church / CPS encodings of the freer monad and freer monad transformer as
-interpreters into a continuation monad `ContT`, an alternative to the inductive
-`Cslib.FreeM`.
+`FreeContT f m` is the Church / CPS encoding of the freer monad transformer
+over an effect signature `f` and base monad `m`. It stores its universal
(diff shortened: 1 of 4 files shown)


quangvdao (approved):
Reviewed with the PolyFun Lean formalization rubric at `44e934f88b32b9a146a4178e0ebb20891a3e47e7`.

No remaining correctness or API blocker. The CPS representation, heterogeneous `liftBind`, lawful base lift, run equations, and one-way `Cslib.FreeM` retraction match the documented contract; the reverse equivalence is correctly not claimed without parametricity.

Review hardening added on this branch:
- response-dependent `liftBind` canaries exercising both Boolean branches;
- base `Option.none` propagation;
- a branch-dependent inductive/CPS roundtrip.

Fresh local gate passed on the exact head: `./scripts/validate.sh --lint --test --axioms` (8,728 declarations / 237 modules, zero `sorryAx` or non-standard axiom taint). Approving pending the refreshed hosted checks.

PR #153, opened 2026-08-23
fix: preserve deprecated tensor simps lemmas

quangvdao:
## Summary

- restore deprecated aliases for the four public simps lemmas generated by the former tensorX and xTensor definitions
- add direct-import compatibility canaries covering all four names against the deprecated expressions
- preserve the y-based API as canonical

## Regression

PR #132 aliased tensorX and xTensor after renaming their definitions, but aliases do not recreate the separately generated tensorX_equivA, tensorX_equivB, xTensor_equivA, and xTensor_equivB declarations. Downstream source using those public theorem names stopped elaborating.

## Validation

- lake build PolyFunTest --wfail (1,322 jobs)
- ./scripts/validate.sh --lint --axioms (library build, module/import/docs checks, environment lint, fixture matrix, zero-taint sweep across 9,013 declarations / 244 modules)
- lake exe lint-style

No mathematical statement or executable behavior changes; this restores only the deprecated source-compatibility surface.

### DIFF
--- a/PolyFun/PFunctor/Equiv/Basic.lean
+++ b/PolyFun/PFunctor/Equiv/Basic.lean
@@ -481,6 +481,10 @@ def yTensor : y ⊗ P ≃ₚ P where
 
 @[deprecated (since := "2026-08-17")] alias tensorX := tensorY
 @[deprecated (since := "2026-08-17")] alias xTensor := yTensor
+@[deprecated (since := "2026-08-17")] alias tensorX_equivA := tensorY_equivA
+@[deprecated (since := "2026-08-17")] alias tensorX_equivB := tensorY_equivB
+@[deprecated (since := "2026-08-17")] alias xTensor_equivA := yTensor_equivA
+@[deprecated (since := "2026-08-17")] alias xTensor_equivB := yTensor_equivB
 
 /-- Tensor product of polynomial functors is commutative up to equivalence -/
 @[simps]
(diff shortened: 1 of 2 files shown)


quangvdao:
## Lean formalization review — `b11f78026748d281304e22f8f6a481107c712a2b`

**Verdict:** no findings; merge once the exact hosted head is green.

### Scope and correctness

- Reproduced the regression across compiled environments: all four `tensorX_*` / `xTensor_*` declarations exist before #132 and are unknown after it.
- The aliases target the canonical `tensorY_*` / `yTensor_*` theorems, so they restore names without duplicating proofs or changing statements, reducibility, executable behavior, universes, or trust assumptions.
- Direct-import canaries exercise every restored name against the corresponding deprecated `tensorX` / `xTensor` expression; this checks the actual downstream elaboration boundary rather than only declaration existence.
- Audited the complete #132 `X`-to-`Y` rename surface. These are the only missing generated `@[simps]` declarations; hand-written compatibility lemmas elsewhere were already aliased.

### Validation

- `lake build PolyFunTest --wfail`: 1,322 jobs passed
- library build: 1,160 jobs passed
- module scope, umbrella imports, docs integrity: passed
- environment and style linters: passed
- axiom-sweep fixture matrix: passed
- project sweep: 9,013 declarations / 244 modules, zero `sorryAx` or non-standard-axiom taint

**Residual risk:** only deprecation-warning behavior for downstream references, which is intentional and matches the existing compatibility policy.

PR #159, opened 2026-08-27
feat(control): indexed support, WriterT support, and the modal-pair bridge

dtumad:
## Summary

This PR completes PolyFun's qualitative support layer around Lean core's `MonadAttach`.
It adds exact indexed support for stateful transformers, an honest `WriterT` instance,
a fuller structural and modal API, and the bridge from free-program weakest
preconditions to the support of interpreted programs.

The change is additive apart from documentation/proof cleanup: 7 files, 928 insertions,
26 deletions.

### Exact and indexed support

- Adds structural equations and automation for `CanReturn` / `support`: pure, bind, map,
  sequence, branching, emptiness, and concrete base-monad unfoldings.
- Adds `StateT.supportFrom` and `ReaderT.supportAt`. Their bind laws keep the same initial
  index through the computation, so they compose exactly even though core's flattened
  transformer support does not.
- Adds state-indexed `AllOutputsFrom`, `SomeOutputFrom`, and `NoOutputFrom`, including
  pure/bind laws and the exact bridge back to flattened value-only guarantees.

### `WriterT`

- Adds `MonadAttach`, `WeaklyLawfulMonadAttach`, `LawfulMonadAttach`, and
  `ExactMonadAttach` for `WriterT`.
- Reachability retains an existential writer accumulator; bind combines witnessed
  accumulators in the transformer's actual multiplication order.

### Modal and WP bridges

- Completes the demonic/angelic `AllOutputs` / `SomeOutput` API, including support
  characterizations, classical duality laws, monotonicity, and the angelic ordered
  monad algebra.
- Adds support transport facts along monad lifts.
- Relates `FreeM.wpVia` to the support of the interpreted program and turns validated
  per-operation specs into `AllOutputs` facts about `liftM`.
- Lowers the assumptions of `ensures_of_allOutputs` and `toWPSound` to the weak
  `MonadAttach` law actually used.

### Tests and documentation

- Pins the `StateT` and `ReaderT` non-exactness boundaries, including a nonempty
  `ReaderT Bool` bind counterexample.
- Pins indexed recovery, the lawful-but-empty `CanReturn` model, WriterT behavior,
  definitional transfer, and a compact set of discriminating `simp` / `grind` canaries.
- Documents `MonadAttach` as the canonical reachability interface and the `SetM` lift
  as a compatibility bridge rather than a new global instance.

## Review follow-up

A full Lean-formalization review found and repaired four issues before merge:

1. The original explanation said `ReaderT` failed only for an empty environment. In
   fact flattened bind composition can fail over a nonempty environment because its two
   premises may choose different readers. The docs are corrected and a concrete
   `ReaderT Bool Id` counterexample now guards this boundary.
2. The indexed-to-flattened `StateT` bridge was described as strictly one-way. For
   postconditions that ignore final state it is an equivalence; both directions and the
   iff theorem are now exposed.
3. One `wpVia` comment incorrectly described a universal handler postcondition as saying
   every response is reachable. It now states the actual typed-output obligation.
4. A newly added angelic-triple theorem was named/documented as a negation of
   `NoOutput` while its statement correctly negated `AllOutputs`. The declaration now
   matches its theorem, and the classical status of the De Morgan laws is explicit.

Redundant duplicate automation examples were also removed; distinct normalization paths
remain covered.

## Validation

At commit `73de11dac10bf7e7fced6f71788c61fce5831d0a`:

  - library build: 1,519 jobs
  - test library: 1,622 jobs
  - axiom sweep: 10,321 declarations across 259 modules
  - zero `sorry` taint and zero nonstandard-axiom taint
- `lake exe lint-style`
- `git diff --check`

All local checks pass.

## Scope

This does not add probability-specific support (`PMF` / `SPMF`), concrete downstream
handler specs, global `WP` instances for `FreeM`, or any CSLib changes.

### DIFF
--- a/PolyFun/Control/Do/Basic.lean
+++ b/PolyFun/Control/Do/Basic.lean
@@ -129,14 +129,26 @@ theorem erasesTo_of_map_eq {α : Type u} {P : α → Prop} {y : m (Subtype P)} {
     (h : Subtype.val <$> y = x) : Std.Do.Internal.ErasesTo y x :=
   ⟨fun {_} k => by rw [← h, bind_map_left]⟩
 
-/-- An "always" judgment yields the `Ensures` witness core's soundness class wants. -/
-theorem ensures_of_allOutputs {α : Type u} {x : m α} {P : α → Prop}
-    (h : AllOutputs P x) : Std.Do.Internal.Ensures P x :=
+omit [ExactMonadAttach m] in
+/-- An "always" judgment yields the `Ensures` witness core's soundness class wants.
+
+Stated at `WeaklyLawfulMonadAttach` rather than `ExactMonadAttach`: the witness is built
+from `attach` and its bind-faithfulness alone, and needs nothing about how possible
+outputs compose. That matters because `StateT`, `ReaderT`, and `EStateM` have the weak
+class but provably not the exact one. -/
+theorem ensures_of_allOutputs {α : Type u} [WeaklyLawfulMonadAttach m] {x : m α}
+    {P : α → Prop} (h : AllOutputs P x) : Std.Do.Internal.Ensures P x :=
   ⟨⟨(fun z : Subtype (CanReturn x) => (⟨z.1, h z.1 z.2⟩ : Subtype P)) <$> MonadAttach.attach x,
     erasesTo_of_map_eq (by rw [Functor.map_map]; exact WeaklyLawfulMonadAttach.map_attach)⟩⟩
 
-/-- The demonic interpretation is sound in core's sense. Not an instance. -/
-theorem toWPSound : letI := toWP (m := m); WPSound m .pure :=
+omit [ExactMonadAttach m] in
+/-- The demonic interpretation is sound in core's sense. Not an instance.
+
+Like `ensures_of_allOutputs` this needs only the weak class, so the *soundness* of the
+demonic reading is available for every monad core equips — including the stateful ones.
+What exactness buys is `toWPMonad`, i.e. the `wp_bind` law, not soundness. -/
+theorem toWPSound [WeaklyLawfulMonadAttach m] :
+    letI := toWP (m := m); WPSound m .pure :=
   letI := toWP (m := m)
   { ensures_of_wp := fun {_ _ _} hwp => ensures_of_allOutputs (hwp True.intro) }
 
--- a/PolyFun/Control/Monad/Support.lean
+++ b/PolyFun/Control/Monad/Support.lean
@@ -7,6 +7,7 @@ module
 
 public import PolyFun.Control.Monad.Algebra
 public import Mathlib.Data.Set.Functor
+public import Mathlib.Control.Monad.Writer
 
 /-!
 # Exact Monadic Support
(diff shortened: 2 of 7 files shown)


dtumad:
Tier 2 and 3 added in `4ff3a2c`. `./scripts/validate.sh --lint --test --axioms` green — **9093 declarations, 0 sorry, 0 non-standard axioms**.

## Per-monad unfoldings

`Id`, `Option`, `Except`, `OptionT`, `ExceptT`, `SetM` now have `canReturn_iff` lemmas and closed-form supports. Their absence was visible in the test file, which worked around it four times with raw `change` and proved `support (none : Option Bool) = ∅` by unfolding `support` and `CanReturn` by hand.

They key on **`CanReturn`, not on membership**. I first wrote them over `a ∈ support x`, and `simpNF` correctly rejected that: `mem_support` already rewrites that redex, so the two race for the same left-hand side. Keyed on `CanReturn` they *extend* the chain — membership → reachability → concrete equation — which is also what the documented normal form says.

## Structural laws

`support_seq`, `support_ite`, `support_dite`, `support_eq_empty_iff`, `support_nonempty_iff`, `not_mem_support_iff`. The last four need no exactness: they are about the shape of the set, not about the operations.

## Transport along a monad lift

Lifting cannot create possible outputs (core's `canReturn_liftM_imp'`), so `support (liftM x) ⊆ support x` and a demonic guarantee transfers for free.

The **introduction half is not available generically, and cannot be**: nothing in `MonadLiftT`'s lawfulness says the lift preserves reachability, and a lift into a monad whose `CanReturn` is uniformly `False` satisfies every law while losing every output. Recorded as a documented gap with `FreeM`'s powerset fold as the worked instance, rather than inventing a class with no consumer yet.

## Judgment theory and angelic symmetry

The pair is now closed under negation both ways, plus disjunction, congruence, and monotonicity **in the computation** — `allOutputs_mono` varied only the predicate, so there was no way to transfer an obligation along a support inclusion.

`mAlgOrderedPropAngelic` had four demonic counterparts and zero theorems of its own; it now has all four.

## The `wpVia` keystone

`wpVia s x post` is by definition `MAlgOrdered.wp (x.liftM s) post`, so under the demonic `Prop` algebra it **is** the "always" judgment on the interpreted program:

```lean
theorem wpVia_demonic_iff_allOutputs (s : Handler n Q) (x : FreeM Q α) (post : α → Prop) :
    wpVia s x post ↔ AllOutputs post (x.liftM s)
```

One rewrite, but it turns `wpFold_le_wpVia` — specs under-approximate the semantic wp — into a statement about which outputs the interpreted program can actually return (`allOutputs_liftM_of_wpFold`). The syntactic half was already connected; this closes the semantic one. Ground direction universe only, because `MAlgOrdered m l` forces `l : Type uB` and `Prop : Type 0`.

## Soundness needs only the weak class — confirmed by building

`ensures_of_allOutputs` and `toWPSound` carried `[ExactMonadAttach m]`, but their witnesses use only `attach` and `WeaklyLawfulMonadAttach.map_attach`. Weakened to `[WeaklyLawfulMonadAttach m]`; it compiles.

The consequence is the point: **`StateT`, `ReaderT`, and `EStateM` have the weak class but provably not the exact one**, so the *soundness* of the demonic reading was available for them all along. What exactness buys is `toWPMonad` — the `wp_bind` law — which is exactly what the indexed bind law in the first commit repairs. The Tier 1 theorem and the Tier 3 theorem are the same theorem.

## Automation

`grind` tags on the 13 directed single-variable bridges only. The support-quantifier characterizations stay `@[simp]`-only: `grind` would case-split the iff, Skolemize the quantifier, and feed the witness back through the always-tagged bind expansions with no finite grounding — the saturation shape VCVio already analysed for `probEvent`. The contract is documented in the module and gated by a battery in `PolyFunTest`, each row closed by a single terminal tactic.

## One deletion

`support_bind_pure_comp` was removed rather than kept: `simpNF` showed `simp` already proves it from `bind_pure_comp` + `support_map`, so it was a duplicate.

PR #162, opened 2026-08-27
feat(uc): add conditional sampler-coherence bridge

dtumad:
## What this adds

This PR strengthens the open-process UC coherence layer from activation equivalence to a strong sampled-path equivalence, conditional on explicit scheduler-transport hypotheses.

- `OpenProcessSamplerEquiv.lean` defines `MonadRelFamily`, exact computation equality (`.eq`), the forgetful relation (`.top`), strong one-step sampler bisimulation, and its equivalence laws.
- `OpenProcessSamplerFactorization.lean` proves sampler-aware plug commutation and the four factorization laws. Each theorem names the scheduler-transport premise it actually needs.
- `SamplerObservation.lean` turns those conditional structural laws into `Observation.RespectsFactorization` for any downstream observation invariant under sampler equivalence.
- The tests exercise the composition bridge and pin the exact-equality obstruction.
- The UC documentation maps the resulting conditional API to the remaining downstream proof obligations.

## Important boundary

This PR exposes a scheduler obstruction; it does not solve it. Reassociating the current shared binary scheduler changes its leaf probabilities. Exact computation equality cannot satisfy both reassociation premises even for a deterministic `Id` scheduler, and equality of ordinary distributional denotations will generally fail for the same reason. A downstream distributional observation therefore still needs different scheduler semantics or an intentionally coarser relation, plus its own adequacy theorem.

Likewise, this is not yet a packet/action security observation. `Observation` compares closed processes, whose external boundary traces are empty; packet/action adequacy remains a downstream obligation.

At `MonadRelFamily.top`, all transport premises are trivial, yielding unconditional strong path matching while deliberately forgetting sampler effects.

## Validation

- `lake exe lint-style`
- `git diff --check`

All pass on the rebased head.


quangvdao:
Thorough Lean review completed and fixes pushed.

Findings:

1. The original description overclaimed readiness for distributional VCVio observations. With the current shared binary scheduler, reassociation changes leaf probabilities, so exact computation equality—and generally equality of distributional denotations—cannot discharge all scheduler-transport premises.
2. The packet-aware wording was also too strong. These observations compare closed processes, where boundary traces are empty; internal packet/action adequacy remains in the downstream invariance theorem.
3. Several tests merely restated theorem signatures rather than exercising behavior.

Changes:

- added `MonadRelFamily.eq` as the strict baseline;
- made scheduler-draw encodings executable and added deterministic evaluation lemmas;
- added a negative canary proving exact equality cannot satisfy both reassociations even for an `Id` scheduler;
- corrected module docs, the UC gate map, roadmap, and PR description;
- removed redundant theorem-restatement tests;
- rebased cleanly onto current `main`.

The core conditional factorization proofs remain intact: their path bijections and the named scheduler-transport hypotheses are coherent. Full build, tests, lint, style, documentation checks, and zero-axiom/zero-`sorry` sweep all pass locally.

PR #168, opened 2026-08-28
docs(reading): re-verify the upstream alignment ledger at the v4.33.1 pin

dtumad:
## Summary

Re-verifies `docs/reading/upstream-alignment.md` row by row at the Lean 4.33.1 pin and against current upstream trees. The previous ledger mixed already-adopted APIs, genuine local abstractions, future Lean 4.35 changes, and stale unused-surface claims.

## Main corrections

- Records completed or partial adoption for the cslib LTS bridge, temporal operators, `Trace.mapHom`, and core list lemmas.
- Reclassifies `MonadHom` as keep-and-bridge rather than adopt; `ApplicativeTransformation` is only its applicative structural analogue.
- Withdraws the unsupported claim that Batteries has an in-flight bundled `MonadHom`.
- Distinguishes the two core weakest-precondition stacks and dates the public `Std.WP`, `LawfulWPMonadAttach`, `mvcgen` deprecation, and stronger coinduction work to Lean 4.35.
- Re-audits the sparse-use table by declarations rather than import paths.
- Records current cslib LTS migration hazards and several upstream APIs worth tracking.

## Review corrections

The final review tightened four boundaries:

1. `Control/Comonad/Instances.lean` has no named in-repo consumer, but is not proved orphaned: it is publicly imported and contributes 65 typeclass instances, whose use is invisible to textual search. Deletion now requires an instance-synthesis and public-API audit.
2. On cslib `main`, `DeterministicStateLabel.finite_image` is added but the existing finite-image instance remains. `MapLabel.lean` is deleted as a module, while `mapLabel` and its main lemmas move into `MapHom.lean`.
3. `ULiftable` is described as a possible universe-transport tool, not a fix for the `MonadAttach (ExceptT ε m)` universe signature; it has no `ExceptT` lifting instance.
4. `CommApplicative` is only an extensional applicative commutation law, not a proof of process independence, fairness, or scheduler invariance. Additional `MonadIter` instances likewise require chosen semantics and law proofs.

## Validation

- rebased onto current `main` after #162
- `python3 scripts/check-docs-integrity.py`
- `git diff --check`

Both local checks pass; the full hosted gate runs on the pushed head.

### DIFF
--- a/docs/reading/upstream-alignment.md
+++ b/docs/reading/upstream-alignment.md
@@ -29,26 +29,35 @@ content, and as little else as it can get away with.**
 
 Every claim below was checked against source on disk, not recalled.
 
-| Tree | Revision surveyed |
-|---|---|
-| Lean core | **v4.33.0** (survey baseline), compared against v4.34.0-rc1 |
-| Mathlib | v4.33.0 (survey baseline), compared against `master` |
-| cslib | v4.33.0 (survey baseline), compared against `main` |
-| Batteries | `main` |
+| Tree | Revision surveyed | Compared against |
+|---|---|---|
+| Lean core | **v4.33.1** | `master` and `v4.34.0-rc2`, via the GitHub API |
+| Mathlib | v4.33.1 (`0df444a360`) | `origin/master` |
+| cslib | v4.33.1 (`98e395a7`) | `origin/main` |
+| Batteries | `4488d40d0` | `origin/main` + live GitHub search |
 
-PolyFun now pins the synchronized v4.33.1 point releases. This comparison was not
-rerun for that toolchain-only update, so the table retains its exact v4.33.0 survey
-baseline.
+Re-run at the v4.33.1 pin (2026-08). The previous baseline was v4.33.0, and the
+earlier hedge that a toolchain-only bump could not have invalidated anything is now
+**verified rather than assumed**: `diff -rq` over the two toolchains' `src/lean`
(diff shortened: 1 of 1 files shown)


quangvdao:
Thorough ledger review completed; fixes are committed and pushed on top of current main.

Actionable findings fixed:

- **Unused-surface methodology:** textual declaration searches cannot prove an instance module unused. `Control/Comonad/Instances.lean` is publicly imported and contributes 65 typeclass instances, so it is now classified as having no named in-repo consumers—not as genuinely orphaned or safe to delete.
- **cslib finite-image claim:** upstream adds `DeterministicStateLabel.finite_image`, but retains the `Finite (lts.image s μ)` instance delegating to it.
- **cslib map-label claim:** the old `MapLabel.lean` module disappears, but `mapLabel`, `mapLabel_tr`, and `mapLabel_mTr` survive in `MapHom.lean`; this is an import-path migration, not removal of the API.
- **Universe transport:** `Mathlib.Control.ULiftable` does not repair core’s `MonadAttach (ExceptT ε m)` signature and has no `ExceptT` instance, so the local alias remains necessary absent a concrete bridge.
- **Abstraction boundaries:** `CommApplicative` does not establish process independence/fairness/scheduler invariance, and new `MonadIter` instances require a chosen semantics plus `LawfulMonadIter` proofs.

The rest of the ledger claims checked against the pinned Lean/Mathlib/cslib sources and current upstream evidence were consistent. Documentation integrity and diff checks pass locally; fresh hosted checks are running.

PR #174, opened 2026-08-29
feat(uc): add reassociation-stable scheduler model

quangvdao:
## Summary

The legacy UC process model installs the same Boolean sampler at every binary composition node. Consequently, a three-component system gives different effective scheduling weights to its leaves depending on whether it is parsed as `(P ∥ Q) ∥ C`, `P ∥ (C ∥ Q)`, or `Q ∥ (C ∥ P)`.

This PR adds the generic, probability-free foundation for fixing that problem:

- `ScheduleMass` records a strictly positive frontier weight and adds under composition.
- `BinaryScheduler` receives the masses of both subtrees instead of blindly flipping the same Boolean sampler.
- `IsFlat` states the key semantic law: every hierarchical three-way draw agrees with one parenthesization-free flat draw; `IsCoherent` derives the swap and reassociation laws needed by UC factorization.
- `ScheduledOpenProcess` preserves the existing binary process representation while carrying total frontier mass and using the mass-aware scheduler at composition nodes.
- A canary demonstrates that merely carrying masses is not enough: a deterministic mass-sensitive policy can still violate reassociation, so the `IsFlat` premise is substantive.

This is deliberately probability-neutral. The next PolyFun slice is #176, and the concrete downstream `ProbComp` scheduler and its exact denotational flatness proof are in Verified-zkEVM/VCVio#572.

## Validation

- zero sorry/axiom debt across the PolyFun namespace


dtumad:
I pushed a review follow-up in 52d85bd.

The main API changes are:

- Replaced the custom `ScheduleMass` structure and its duplicated instances/lemmas with Mathlib's standard positive naturals, `PNat` (`ℕ+`). This keeps positivity intrinsic while reusing the existing `One`, additive, order, coercion, and simp APIs.
- Kept `ScheduledOpenProcess.withMass`, `.atom`, `.mapBoundary`, and `.interleave` opaque, but added public `[simp]` equations characterizing their `mass` and `process` projections. Together with the existing extensionality theorem, this exposes the useful API without publishing the implementation bodies.
- Exposed only `scheduledOpenTheory`, since its `Obj` family needs to reduce for ordinary downstream `OpenTheory` expressions to elaborate ergonomically, matching the existing `openTheory` precedent.
- Added ordinary-import canaries for these equations and the theory object family, and corrected the documented scheduler dependency edge.

Validation: `./scripts/validate.sh --lint --test --axioms` passes, including zero axiom/sorry debt.

PR #176, opened 2026-08-29
feat(uc): lift scheduled draws through factorization

quangvdao:
## Summary

Stacked on #174.

This PR connects the mass-aware scheduler algebra to the existing UC factorization path maps. For arbitrary component samplers, it proves that:

- the source-shaped nested sampler and the left-reassociated sampler are related after `parLeftPathEquiv`;
- the source-shaped nested sampler and the right-reassociated sampler are related after `parRightPathEquiv`.

Both results consume only `BinaryScheduler.IsCoherent`. Their proofs flatten the nested Boolean draws to a common leaf continuation and then use the scheduler law under `MonadRelFamily.bind_congr`. No probability assumptions enter PolyFun.

The three `*_bind` lemmas expose the precise nested scheduler calls and make the bridge reusable by downstream process- and observation-level theorems. The concrete downstream consumer is Verified-zkEVM/VCVio#572, which pins this PR's exact head and proves coherence for proportional `ProbComp` scheduling.

## Validation

- zero sorry/axiom debt across the PolyFun namespace


dtumad:
I restacked this PR on the updated scheduler branch and pushed the follow-up in 08a18fa.

- Updated all public scheduled-factorization theorem signatures to use Mathlib's `PNat` (`ℕ+`) rather than the removed custom `ScheduleMass` type.
- Added an ordinary-import canary for the public `BinaryScheduler.sourceDraw_bind` API. This module only introduces public theorems, so it does not need any additional `@[expose]` declarations.
- Added the new module to the interaction wiki map and tightened the scheduler documentation.

The full stacked tree passes `./scripts/validate.sh --lint --test --axioms`, including zero axiom/sorry debt.

PR #178, opened 2026-08-29
feat(tocslib): add polynomial-time single-tape toolkit

quangvdao:
## Summary

Adds a separate `ToCslib` Lake library for reusable extensions of the pinned cslib single-tape machine API. This is the low-level first slice extracted from VCVio #500.

- encoded polynomial-time witnesses with explicit machine-description size;
- fixed-width and variable-width binary encoding families, with independent source/target universes;
- concrete constant and finite-table machines;
- canonical finite machine tables and the counting/diagonalization toolkit;
- direct producer canaries for bit order, finite-table semantics, bit overwrite, and function-space cardinality.

## Ownership

`ToCslib` imports only cslib and Mathlib. It does not import PolyFun realizability, oracle semantics, probability, or cryptography. The later PolyFun backend adapter will import it explicitly; the generated `PolyFun` umbrella does not.

## Validation

- `lake build ToCslib`
- `lake build PolyFunTest.ToCslib.Basic`
- `lake build PolyFun`
- `lake lint` (now covers both `PolyFun` and `ToCslib`)
- `./scripts/validate.sh --lint --test --axioms`
- `git diff --check`

All pass locally on Lean 4.33.1, with zero sorry/axiom debt under `PolyFun` and `ToCslib`.

## Provenance

The state-normalization and finite-machine covering proofs, together with the fixed-bit append machine, include the proof work contributed by Elias Judin in [VCVio #487](https://github.com/Verified-zkEVM/VCVio/pull/487), carried through [VCVio #500](https://github.com/Verified-zkEVM/VCVio/pull/500). The original commit records `Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>`. The current `ToCslib/Computability/SingleTape/Counting.lean` and `Snoc.lean` headers retain Elias Judin's authorship.

PR #179, opened 2026-08-29
feat(realizability): bridge local nonuniform machine certificates

quangvdao:
Adds an optional bridge from PolyFun programs to cslib machine families with fixed encodings, polynomial resource bounds, and bounded semantic implementation. Concrete machine and complexity theory remains local in `ToCslib` while upstream APIs stabilize; `PolyFunCslib` contains the PolyFun-specific certificate and bridge.

- Certify initialization, combined return/query observation, and partial updates with `EncPolyTimeFam`. Bound time envelopes, descriptions, hidden-state encoding lengths, and query rounds independently.
- Require progress at every reachable query, including rejection of empty-answer stuck programs. Close certificates under pointwise equality and code-supplied input/result maps.
- Reuse generic quantitative traces. `Witness.executionWork_le_totalTime` bounds every prefix by `initTime + (rounds + 1) * headTime + rounds * updateTime`, deriving its round bound from implementation. The reusable trace length and additive work lemmas live in `PolyFun/Realizability/Quantitative.lean`.
- Prove `ToCslib.Computability.exists_not_realizableLE_poly` entirely in the local machine layer: some Boolean predicate family has no polynomial bound on its realizing machine-pair descriptions. The optional adapter derives `exists_not_isPPolyBy_pure` by extracting the initial-state initialization/observation pair from a pure-program certificate.
- Cover the optional production library in the existing build, lint, style, module, documentation-integrity, and axiom checks. Add no scripts. Keep `PolyFun` as the default target and exclude the concrete libraries from its generated umbrella.

The cost is a certified envelope for local cslib machines. It excludes external-answer computation and does not establish exact linked whole-program cost or equivalence with the textbook circuit definition of P/poly. Boundary representations are fixed by the caller; closure requires actual supplied code. Scope and source comments state these limits explicitly.

Validation: `./scripts/validate.sh --lint --test --axioms` and `git diff --check` pass on the unchanged Lean 4.33.1 dependency pins. The production sweep checks 10,997 declarations across 284 modules with zero sorry/nonstandard-axiom taint. Explicit axiom checks pass for both separation theorems, the witness cost bound, closure, and generic trace lemmas. Ordinary-import examples cover pure returns, real queries and distinct answers, mismatched indices, nontrivial maps, empty-answer rejection, and the lower-layer separation independently of PolyFun. Documentation-checker fixtures cover the auxiliary roots.

## Provenance

This adapter uses the machine-counting development in [PolyFun #178](https://github.com/Verified-zkEVM/PolyFun/pull/178), including the state-normalization and covering proofs contributed by Elias Judin in [VCVio #487](https://github.com/Verified-zkEVM/VCVio/pull/487) and ported through [VCVio #500](https://github.com/Verified-zkEVM/VCVio/pull/500). That original commit records `Co-authored-by: Aristotle (Harmonic) <aristotle-harmonic@harmonic.fun>`. The current pure-program bridge uses certified initialization and initial-head observation; `realizableLE_of_isPPolyBy_pure` and `exists_not_isPPolyBy_pure` provide the corresponding nontriviality endpoint for the new model.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
         run: lake exe cache get
 
       - name: Build production libraries
-        run: lake build PolyFun ToCslib --wfail
+        run: lake build PolyFun ToCslib PolyFunCslib --wfail
 
       - name: Run environment linters
         run: lake lint
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -20,6 +20,6 @@ jobs:
         with:
           mode: check
       # The action invokes lint-style with the default target (PolyFun).
-      # ToCslib is intentionally a separate, non-default production library.
-      - name: Lint auxiliary production library
-        run: lake exe lint-style ToCslib
+      # ToCslib and PolyFunCslib are separate, non-default production libraries.
+      - name: Lint auxiliary production libraries
+        run: lake exe lint-style ToCslib PolyFunCslib
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -101,8 +101,8 @@ and depend on this library.
   admissible translation, boundary-level realizability invariance, and
   admissible word encode/decode retractions. Generic quantitative realizability
   and backend-relative trace accounting also live here. Optional concrete
-  realizability adapters are isolated under `Realizability/Backend/`; the cslib
-  P/poly adapter exposes a non-uniform, boundary-pinned certificate without
+  realizability adapters live in separate library roots; `PolyFunCslib/` exposes
+  a non-uniform, boundary-pinned P/poly certificate without
   importing oracle, probability, or cryptographic policy. Named cryptographic
   adversary classes and protocol-specific adequacy results remain downstream.
 - `PolyFun/Control/`: monad and comonad infrastructure transitively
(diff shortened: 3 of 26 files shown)


dtumad:
I reviewed the backend and P/poly development and pushed 8205bbb with the review adjustments. This branch was rebased onto the updated #178 head, so the update used force-with-lease against the previously verified remote head.

The patch replaces proof-oriented reducer exposure with modular APIs: opaque Boundary constructors and conversion with projection laws, public ProgramProgress equations, an Implements characterization plus intro/apply lemmas, IsPPolyBy intro/elimination lemmas, and the RealizableLE membership characterization. The ordinary-import tests now construct and consume certificates through those APIs.

Three exposed definitions remain intentionally. encodingStepClass and quantitative are reducibly exposed because their Str and Realizer type families must elaborate respectively as raw Boolean encodings and EncPolyTime across module boundaries. Realization.toQuantitative remains exposed because the generic ExecutionTrace API is dependently indexed by the adapted machine-state type; named theorems cannot repair that equality after elaboration. Each contract is now documented next to the definition, while named projection and cost laws form the proof API.

I also checked the backend cost interpretation, boundary encodings, progress condition, trace/time/description bounds, closure constructions, and the pure Boolean counting/diagonal nontriviality bridge. I did not find a mathematical blocker.

The stacked checkout passes ./scripts/validate.sh --lint --test --axioms, including PolyFunCslib, its ordinary-import canaries, and the zero axiom/sorry sweep. GitHub CI has started on the updated head.

PR #184, opened 2026-09-03
chore(deps): bump Lean, Mathlib, and cslib to v4.34.0-rc2

dtumad:
## Summary

Moves the toolchain pin from v4.33.1 to v4.34.0-rc2, with Mathlib and cslib at their matching rc2 tags and the Lake manifest regenerated. This is the prerequisite for building the program-logic layer on core's lattice-generic `Std.Internal.Do` stack (`vcgen`): at v4.33.1 its triple notation cannot even synthesize `CCPO EPost⟨⟩` without `open Lean.Order`; rc2 adds the scoped instance.

Source adjustments the new pins require:

- core deprecates `if_pos` / `if_neg` / `dif_pos` / `dif_neg` in favour of the statement-identical `ite_eq_left` / `ite_eq_right` / `dite_eq_left` / `dite_eq_right`; every use in `PolyFun`, `PolyFunTest`, and `ToCslib` is renamed so `--wfail` stays clean;
- cslib's `IsSimulation`, `IsBisimulation`, and `IsSWBisimulation` now bind their states strictly implicitly; the two explicit-state application sites in `Control/Bisimulation.lean` drop those arguments;
- `repeatM_eq_of_monadTail` is now `repeatM.Internal.eq_of_monadTail` (docstring references only).

Policy and docs:

- `scripts/check-modules.sh` now also matches `Std.Internal.Do`, which core ships alongside `Std.Do` and which the previous regex let through; AGENTS.md gotcha 8 and the wiki say so.
- `docs/reading/upstream-alignment.md` is re-surveyed at the rc2 pin: real diff stats, a corrected `vcgen` row (`vcgen` consumes `Std.Internal.Do.WPMonad`, not `Std.Do.WP`, so retargeting needs new instances rather than a bump), and the cslib LTS breaking changes split into the two that landed at the pin and the `MapLabel` removal still ahead of it.

Note: merging this changes `lean-toolchain` on `main`, so `release-tag.yml` will cut a `v4.34.0-rc2` release tag; downstream pins (VCVio, ArkLib) can target it when they bump.

## Validation

- `lake build --wfail`, `lake build PolyFunTest --wfail`, `lake build ToCslib --wfail`
- `./scripts/validate.sh --lint --test --axioms` (lint passes for `PolyFun` and `ToCslib`; zero sorry/axiom taint across 10903 declarations)
- `./scripts/check-modules.sh` self-test with the extended `Std.Internal.Do` matcher

https://claude.ai/code/session_01Nb4g3Vxgjhnhsca7BzAsFc


dtumad:
Marking draft until an actual non-candidate release

Issue #244, opened 2026-09-19
ComplexityBackends: generalize the generic core of the cslib backend and survey further backends


dtumad:
Follow-up to the `ComplexityBackends` restructure in #243. This is the epic for generalizing the generic core of the cslib backend, surveying further backends, and the policy investigations that came out of the exploration. Report: https://claude.ai/artifact/BmKjsFVj1YcumBqr87hnvk (sorry-free spikes over three backends, an exposure audit, a literature map; summary in the comment below).

## Verdicts (from the exploration)

| Unit | Verdict |
|---|---|
| Counting separation (`RealizableLE`, cover, diagonal, `exists_not_realizableLE_poly`) | generic: needs only a `DescriptionMeasure` (description size, boundary-indexed finite canonical descriptions, "equal descriptions against a faithful codomain compute the same function"); no cost, time or category enters |
| Machine residue for cslib | state relabelling, run determinism, the output law of a witness, the table count |
| `EncPolyTimeFam` composition, `const`, `ofFintype` | generic over a `PolynomialBackend` (canonical per-code polynomials with envelope/overhead laws) and a `FiniteTables` mixin; `EncPolyTimeFam` is the generic `FamRealizer` field for field. Families over per-use `PolyRealizer` certificates cannot compose |
| `PPoly.IsPPolyBy` (A) vs core `PolynomialProgramWitness` (B) vs VCVio's uniform `SecurityFamily.IsOraclePPTBy` (C) | (A) ⇒ (B) at each `n` with constant polynomials; (C) ⇒ (A) by recoding at the unary-parameter boundary; uniform non-vacuity holds; the `1^n` convention is `time.eval (n + k)` at the family layer and the unary parameter in the packed input at the uniform layer |
| Module system | the backend is a minor consumer of core exposure; `Closure.lean` narrows with five `@[expose]`s; `import all` never rescues definitions; exported `rfl` lemmas over un-exposed defs fail in the defining module |
| Second backends | complexitylib circuits (exact zero-overhead category, pairing, sums) and multi-tape machines both instantiate the same axiom set; feedback: `Desc` must be boundary-indexed, tape count enters `descSize` exponentially |

## PR sequence (native stack on #243)

- [x] PR 0 #245 — `import all` boundaries in `scripts/check-modules.sh`, the exposure-policy table and the two module-system facts in `docs/development/module-api.md`.
[...]

PR #249, opened 2026-09-19
feat(realizability): polynomial backends and uniform realizer families

dtumad:
Draft, stacked on #248 (`baseline/description-measures`). Part of #244 (PR 2 of the baseline sequence).

## What this adds

**Generic (`PolyFun/Realizability/Quantitative/Family.lean`)**
- `DescriptionMeasure.PolynomialBackend`: a backend whose realizers carry a canonical polynomial certificate `timeOf`, with an output-size `envelope`, a composition `overhead` (both functions of the certificates), uniform identity bounds, and subadditive description size. Every law is in the shifted form `p.eval k ≤ P.eval (n + k)` that families consume.
- `FamRealizer PB a b f`: one realizer per parameter `n`, a uniform time polynomial in `n + k` (`k` the encoded input size) and a uniform description polynomial in `n` (the advice bound). `cost_le`, `id`, `comp`, with projection laws by `rfl`.
- `FiniteTables`: the advice primitive (linear time, table-sized description); `FamRealizer.ofFintype` builds a family from polynomially small finite domains.
- `FamRealizer.mem_realizableLE`: a family's advice bound is a description bound at every parameter. `exists_not_famRealizer` uses it to transport #248's counting separation to this layer — `IsEmpty (FamRealizer PB a b (fun n ↦ ι n ∘ f n))` for some Boolean predicate family. Without this bridge the uniform-family layer and the `RealizableLE` counting layer are not known to be talking about the same functions.

**Backend (`ComplexityBackends/CslibSingleTape/Family.lean`)**
- `Backend.polynomialBackend`: certified time as `timeOf`, envelope `1 + X + p`, overhead `0`, from `EncPolyTime.length_le` / `comp_time_eval` / `size_comp`. `comp_time_eval` gives `(h.comp h').time.eval k = h.time.eval k + h'.time.eval (1 + k + h.time.eval k)`, which is exactly `timeOf_compose_le`'s right-hand side at that envelope, so this backend has no composition overhead; the earlier `q.comp (1 + X + p)` counted the second machine twice.
- `Backend.finiteTables` from `EncPolyTime.ofFintype` and its bounds.
- `EncPolyTimeFam.toFam` / `ofFam`: the existing uniform family notion is the generic one field for field, inverse by `rfl`.

**Tests**: `PolyFunTest/Realizability/QuantitativeFamily.lean` (degenerate all-zero instance over the cost-free backend, so the generic layer carries no complexity content of its own), `PolyFunTest/ComplexityBackends/CslibSingleTape/Family.lean` (round trips, composition laws, table = table machine), and the ordinary-import canary `ModuleAPI/Realizability.lean`.

**Docs**: guide section "Polynomial backends and families" in `docs/guides/realizability.md`; `REFERENCES.md` gains `DH11` (Dal Lago–Hofmann realizability with resource majorants) and `Cob65` (Cobham's function algebra; `FiniteTables`/`id`/`comp` are its initial-function and composition clauses, limited recursion has no generic counterpart yet).

## Design notes

- Why canonical certificates and not `PolyRealizer`: composition overhead on the single-tape backend is the second machine's polynomial at the first machine's output envelope, a hypothetical length that a per-input certificate does not control. Families of `PolyRealizer`s do not compose there.
- The `n + k` bound form is the formal content of the `1^n` convention at this layer.
- Generic composition folds the parameter into the envelope's argument, so the composite bound is one substitution looser than the backend's own `EncPolyTimeFam.comp`. Both are recorded (`time_comp` test).
- Exposure: `FamRealizer.id`/`comp`/`ofFintype`, `Backend.polynomialBackend`, `Backend.finiteTables`, `toFam`/`ofFam` are `@[expose]` because their projection and round-trip laws are stated as exported `rfl` theorems (the module system rejects exported `rfl` over unexposed bodies). `trivialMeasure` in the description test became `@[expose]` for the same reason.

## Adversarial notes (per #244 §E)

Unchanged threat model: a certificate can only overstate cost; `PolynomialBackend` laws are inequalities, so a backend cannot claim a smaller polynomial than its cost. `FiniteTables` bounds the advice by the table size, never below it. Open: no whole-program linking yet (PR 3), and `Type 0` pin on the backend.

## Validation

`./scripts/validate.sh --lint --test --axioms` green; no `sorry`.


quangvdao:
Two structural follow-ups from reviewing this PR. Neither blocks it; both are bigger than it.

### 1. `PolynomialBackend` bundles two independent contracts

Of the 13 fields, ten never mention `M`: `timeOf`, `cost_le`, `envelope`, `size_le`, `envelope_le`, `overhead`, `overhead_le`, `timeOf_compose_le`, `idTime`, `timeOf_identity_le`. The other three — `idDesc`, `descSize_identity_le`, `descSize_compose_le` — are the only reason the structure is parameterized by a `DescriptionMeasure` at all.

So a backend that certifies time but has no advice notion, or that takes description size from a different measure, still has to produce `idDesc` and `descSize_compose_le` before it can reach the time laws. Splitting into an `M`-free core plus a `DescriptionSize M` mixin, with `PolynomialBackend` extending both, leaves `FamRealizer` unchanged (it needs both) and only splits the instantiation obligation; `Backend.polynomialBackend` becomes two `where` blocks.

Cheap now, awkward later: the first backend that isn't `CslibSingleTape` is where the merged bundle actually costs something.

### 2. `FamRealizer` has `id` and `comp` but no laws — and cannot have them yet

There is no `id_comp`, `comp_id` or `comp_assoc`, and no `Equiv` packaging. That is not an omission a few `rfl`s would close. It is blocked at two independent levels.

**`HasCategory` has no category laws.** `PolyFun/Realizability/Quantitative.lean:83` declares `identity`, `compose`, `composeOverhead`, `cost_compose_le`, and nothing else — no associativity, no unit law. So `(X.comp Y).comp Z` and `X.comp (Y.comp Z)` are not known to agree even on `wit`. (The class name overstates what it provides: it is a composition structure with a cost bound, not a category.)

**The `time` fields genuinely differ.** Writing `E := PB.envelope`, `O := PB.overhead`:

```
(X.comp Y).comp Z |>.time = Txy + Tz.comp (X + E Txy) + O Txy Tz    where Txy = Tx + Ty.comp (X + E Tx) + O Tx Ty
X.comp (Y.comp Z) |>.time = Tx  + Tyz.comp (X + E Tx) + O Tx Tyz    where Tyz = Ty + Tz.comp (X + E Ty) + O Ty Tz
```

At this PR's single-tape instance (`E p = 1 + X + p`, `O = 0`), take `Tx = Ty = 0` and `Tz = X`. Left-associated: `Tz.comp (1 + 2X + Txy) = 1 + 2X`. Right-associated: `Tz.comp ((1 + 2X + Ty).comp (1 + 2X + Tx)) = 3 + 4X`. Hand computation from the definitions, not checked in Lean.

So the layer needs a chosen notion of equivalence before any law can be stated. Options I can see:

- mutual domination of `time` and `desc` with `wit` equality — but `wit` equality still needs `HasCategory` laws;
- forget the certificates and remember only the realized function family — the laws become trivial and the layer loses its content;
- treat the substitution as a 2-cell and present this bicategorically.

I do not have a recommendation. The right choice depends on what PR 3's linking theorem needs to transport across composition, so it is better decided there than here.

### Aside, pre-existing

`ComplexityBackends/CslibSingleTape/Backend.lean:103` sets `HasCategory.composeOverhead` to the full `second.time.eval (1 + |x| + first.time.eval |x|)`, while the shape of `cost_compose_le` already charges `Q.cost second (f input)` separately. That bound therefore counts the second machine twice. It is unrelated to `PolynomialBackend.overhead`, which this PR correctly sets to `0` — noting it only because the two overheads are easy to confuse.

